PR #8, opened 2015-02-02
Whitelist.all() should be scoped per Jenkins instance

jglick:
I found that when two or more functional test cases share a single JVM (for example, live in the same suite), static store of `Whitelist.all()` could prevent subsequent tests from running correctly if they made use of `ScriptApproval.approvedSignatures` via `ApprovedWhitelist`.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
@@ -29,6 +29,8 @@ import hudson.ExtensionPoint;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.WeakHashMap;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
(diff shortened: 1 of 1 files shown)


tfennelly:
:+1:

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
```diff
@@ -69,12 +71,18 @@
...
         if (all == null) {
             // TODO should check for dynamic changes in this list, e.g. from dynamically loaded plugins, and return a fresh aggregate
-            all = new ProxyWhitelist(Jenkins.getInstance().getExtensionList(Whitelist.class));
+            all = new ProxyWhitelist(j.getExtensionList(Whitelist.class));
+            allByJenkins.put(j, all);
         }
         return all;
     }
-    private static Whitelist all;
+    private static final Map<Jenkins,Whitelist> allByJenkins = new WeakHashMap<Jenkins,Whitelist>();
```
recampbell: Curious why you didn't use Jenkins.lookup here.

recampbell:
:+1:

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
jglick: That is what I tried first, but it did not work. I had an `@Extension` which was `AllHolder` with a field corresponding to the `ProxyWhitelist`, but neither `Jenkins.getInstance().getExtensionList(AllHolder.class).get(AllHolder.class)` nor `Jenkins.lookup(AllHolder.class)` returned anything; perhaps because `AllHolder` was not extending any interface so it was not being categorized as anything.

I also considered registering a `ProxyWhitelist` subclass, which solves that problem, but then introduces a different problem that the extension list now includes itself, which I would need to filter out.

So just using a `WeakHashMap` seemed simpler than fighting these issues.

PR #12, opened 2015-04-29
More generic methods whitelisted

olivergondza:
Whitelist some more prototypes I run into.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -11,7 +11,21 @@ method java.util.Collection size
 method java.util.Map put java.lang.Object java.lang.Object
 method java.util.List get int
 new java.net.URL java.lang.String
+method java.lang.String contains java.lang.CharSequence
+method java.lang.String endsWith java.lang.String
+method java.lang.String indexOf java.lang.String
+method java.lang.String toLowerCase
+method java.lang.String matches java.lang.String
 method java.lang.String split java.lang.String
-staticMethod org.codehaus.groovy.runtime.ScriptBytecodeAdapter findRegex java.lang.Object java.lang.Object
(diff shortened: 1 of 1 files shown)


olivergondza:
@jglick, can I start throwing more harmless methods in, or this really needs to be reviewed?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -11,7 +11,22 @@ method java.util.Collection size
...
+method java.lang.String contains java.lang.CharSequence
+method java.lang.String endsWith java.lang.String
+method java.lang.String indexOf java.lang.String
+method java.lang.String toLowerCase
+method java.lang.String matches java.lang.String
 method java.lang.String split java.lang.String
+method java.lang.String startsWith java.lang.String
+method java.lang.String replace java.lang.CharSequence java.lang.CharSequence
+method java.util.Collection contains java.lang.Object
+staticMethod java.lang.System getenv java.lang.String
```
jglick: I would not add `getenv` as it could perhaps be used to steal secrets from the master. (Probably not, but seems suspicious.)

jglick:
It needs to be reviewed I think, but yes please go ahead and add more to the PR if you have the time.

BTW [JENKINS-25804](https://issues.jenkins-ci.org/browse/JENKINS-25804).

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
olivergondza: I found it in couple of our combination filters so I am interested in this one. It i imho impossible to keep the env-vars private if we allow master to run builds (even matrix parents). Perhaps it is not even necessary.

jglick: > if we allow master to run builds

If you allow that, then you might as well disable security entirely.

olivergondza:
Removing `System#getenv()` so we can keep moving.

jglick:
:+1:, feel free to merge or continue.

PR #13, opened 2015-05-27
[FIXED JENKINS-28586] handle method calls/property accesses via closure

kohsuke:
[JENKINS-28586](https://issues.jenkins-ci.org/browse/JENKINS-28586)

This PR incorporates a new version of groovy-sandbox that understands how `Closure` proxies method calls and property accesses to a configurable combination of `owner` and `delegate`. It turns out this was needed to make jenkinsci/workflow-plugin #131 work correctly in some of its advanced use cases.

The commit history retains my earlier attempt to fix the problem in this plugin, but later I moved the fix over to groovy-sandbox.

Also relevant: kohsuke/groovy-sandbox #11


kohsuke:
[The actual fix in groovy-sandbox](https://github.com/kohsuke/groovy-sandbox/commit/74854025db69cc12e69a8ffea9566222f8e6842b). I've already released this so as not to block you while I'm away, but @jglick should probably look at this change.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -29,3 +29,6 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods disjoint java.util
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods join java.util.Collection java.lang.String
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods or java.lang.Boolean java.lang.Boolean
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods xor java.lang.Boolean java.lang.Boolean
+staticField groovy.lang.Closure *
+field groovy.lang.Closure resolveStrategy
```
jglick: Should rather whitelist the getters and setters.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -29,3 +29,6 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods disjoint java.util
...
+field groovy.lang.Closure delegate
```
jglick: Ditto.

jglick:
@reviewbybees I think this is ready now.

rsandell:
:+1: As far as I can grasp the implications.

kohsuke:
:+1:

PR #15, opened 2015-05-28
Some more useful things to whitelist

jglick:
Taken from [`DockerDSL`](https://github.com/jenkinsci/docker-workflow-plugin/blob/1beac603eea45e8f44479fe992b044ba862b277d/src/main/java/org/jenkinsci/plugins/docker/workflow/DockerDSL.java#L53-65).

@reviewbybees


tfennelly:
:+1:

jtnord:
:+1:

PR #16, opened 2015-06-17
[WiP] Handle closure receiver

jglick:
Needed to solve some, but not all, test failures from https://github.com/jenkinsci/workflow-plugin/pull/131.

@reviewbybees esp. @kohsuke

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
 
+import groovy.lang.Closure;
 import groovy.lang.MissingPropertyException;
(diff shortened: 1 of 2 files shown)


tfennelly:
LGTM from just a code pov, but probably needs a @kohsuke review too.

:+1:

kohsuke:
It could be my jetlag, but I thought I made [the change in groovy-sandbox](https://github.com/kohsuke/groovy-sandbox/commit/74854025db69cc12e69a8ffea9566222f8e6842b) so that this kind of change in `GroovyInterceptor` impls is unnecessary.

That is, when Groovy executes code that calls into a method on `Closure`, `groovy-sandbox` recognizes that, deciphers where the method call is ultimately dispatched to (using the proper resolution strategy), and then call `GroovyInterceptor` methods accordingly.

Sounds like that's not happening. I think there's a disconnect.

jglick:
Sounds like `CpsClosure2` fixes this in https://github.com/jenkinsci/workflow-plugin/pull/131, meaning the problem was not in `groovy-sandbox` per se but `groovy-cps`, so I hope this patch is unnecessary; closing unless I hear otherwise.

PR #17, opened 2015-06-19
[JENKINS-28154] Reproduce in unittest

olivergondza:
[JENKINS-28154](https://issues.jenkins-ci.org/browse/JENKINS-28154)

I tried to exercise as much operators/keywords as possible, though only `in` (membership operator, not foreach separator) seems to cause this problem. I also whitelisted  some methods.


olivergondza:
https://github.com/kohsuke/groovy-sandbox/pull/26

jglick:
If you mark this

```java
@Ignore("TODO JENKINS-28154")
```

I would consider it mergeable.

olivergondza:
It can be merged as is after https://github.com/kohsuke/groovy-sandbox/pull/26 is integrated :wink:.

jglick:
Once #22 is merged, you should merge with `master` and this will get rechecked.

jglick:
BTW it was not necessary to close this, as when #23 is merged, GH will automatically display this as merged as well (since the branch head would be an ancestor of the `master` branch head).

PR #18, opened 2015-07-17
Enable FindBugs runs on "verify" stages + XML report generation

oleg-nenashev:
- [x] Reconfigure pom.xml
- [x] Enable FindBugs on https://jenkins.ci.cloudbees.com/job/plugins/job/ssh-slaves-plugin/
- [x] Fix important issues

The code is not clean according to the local FindBugs run

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
@@ -360,7 +360,7 @@ import org.kohsuke.stapler.bind.JavaScriptMethod;
                 throw new AssertionError(x);
             } finally {
                 if (input != null) {
-                    input.close();;
+                    input.close();
                 }
             }
         } finally {
@@ -433,7 +433,7 @@ import org.kohsuke.stapler.bind.JavaScriptMethod;
     /**
      * Called when configuring a classpath entry.
      * Usage is similar to {@link #configuring(String, Language, ApprovalContext)}.
-     * @param entry entry, which is being configured
+     * @param entry entry to be configured
      * @param context any additional information
      * @throws IllegalStateException {@link Jenkins} instance is not ready
      */
@@ -840,6 +840,7 @@ import org.kohsuke.stapler.bind.JavaScriptMethod;
         return getClasspathRenderInfo();
     }
 
+    //TODO: Remove once the baseline supports Jenkins.getActiveInstance (1.590+)
     @Nonnull
     private static Jenkins getJenkins() throws IOException {
         final Jenkins jenkins = Jenkins.getInstance();
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
@@ -45,6 +45,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
@@ -59,6 +60,7 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import net.sf.json.JSON;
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
@reviewbybees
Fixed several issues, got timeboxed on others

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -334,6 +358,10 @@ private static String hashClasspathEntry(URL entry) throws IOException {
                 return Util.toHexString(digest.digest());
             } catch (NoSuchAlgorithmException x) {
                 throw new AssertionError(x);
+            } finally {
+                if (input != null) {
+                    input.close();;
```
jglick: Unnecessary, this is only a wrapper stream, the base stream is only being closed, FindBugs is being dumb, suppress it.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -403,10 +433,13 @@ public synchronized String using(@Nonnull String script, @Nonnull Language langu
     /**
      * Called when configuring a classpath entry.
      * Usage is similar to {@link #configuring(String, Language, ApprovalContext)}.
-     * @param url the location of the entry
+     * @param entry entry, which is being configured
```
jglick: BTW incorrect use of English commas.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -778,10 +833,28 @@ public JSON denyApprovedClasspathEntry(String hash) throws IOException {
...
+    private static Jenkins getJenkins() throws IOException {
+        final Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins == null) {
+            throw new IOException("Jenkins instance is not ready");
+        }
+        return jenkins;
+    }
+    
+    @Nonnull
+    private static Jenkins getActiveJenkinsInstance() throws IllegalStateException {
```
jglick: Should have a `TODO` comment to remove once the baseline supports `Jenkins.getActiveInstance`.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
oleg-nenashev: I cannot suppress a single call, but suppressing the entire method is a hazardous thing. I would keep it as is (with a comment). Do you agree?

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
oleg-nenashev: It throws IOException here, but probably it's another case to Jenkins core

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
oleg-nenashev: Uhm, entry.openStream may open a FileInputStream. I think that the FindBugs complain was valid

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -334,6 +358,10 @@ private static String hashClasspathEntry(URL entry) throws IOException {
                 return Util.toHexString(digest.digest());
             } catch (NoSuchAlgorithmException x) {
                 throw new AssertionError(x);
+            } finally {
+                if (input != null) {
+                    input.close();
+                }
```
jglick: BTW the simpler idiom (all local variables effectively final) is

```java
MessageDigest digest;
try {
    digest = MessageDigest.getInstance("SHA-1");
} catch (NoSuchAlgorithmException x) {
    throw new AssertionError(x);
}
DigestInputStream input = new DigestInputStream(new BufferedInputStream(is), digest);
try {
    // …
} finally {
    input.close();
}
```

which can also be mechanically converted more easily to a `-source 7` `try`-with-resource block.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
jglick: Not sure I understood that comment. Anyway I still request a

```java
// TODO 1.590+ use Jenkins.getActiveInstance
```

jglick:
:bee:

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
jglick: Never mind, you just put the comment above the other method.

PR #19, opened 2015-07-17
Fixed incorrect sandbox check when invoking GDM method on a class that defines invokeMethod.

kohsuke:
The groovy runtime would resolve a call to a GroovyDefaultMethods method in this case, but the script-security check incorrectly concludes that the call would resolve to the invokeMethod method.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -44,3 +44,4 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods disjoint java.util
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods join java.util.Collection java.lang.String
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods or java.lang.Boolean java.lang.Boolean
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods xor java.lang.Boolean java.lang.Boolean
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods asBoolean java.lang.Object
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
@@ -26,16 +26,19 @@ package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
 
 import groovy.json.JsonBuilder;
 import groovy.json.JsonDelegate;
+import groovy.lang.Closure;
 import groovy.lang.GString;
 import groovy.lang.GroovyObject;
 import groovy.lang.GroovyObjectSupport;
 import groovy.lang.GroovyShell;
+import groovy.lang.MetaMethod;
 import groovy.lang.MissingPropertyException;
 import groovy.lang.Script;
 import groovy.text.SimpleTemplateEngine;
 import groovy.text.Template;
 import hudson.Functions;
 
+import java.lang.reflect.Constructor;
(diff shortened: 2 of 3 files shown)


kohsuke:
@reviewbybees

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -44,3 +44,4 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods disjoint java.util
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods join java.util.Collection java.lang.String
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods or java.lang.Boolean java.lang.Boolean
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods xor java.lang.Boolean java.lang.Boolean
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods asBoolean java.lang.Object
```
amuniz: This is added as a side change, not related with the PR subject, right?

Note: I'm only trying to understand how this plugin works, so my comments are not aiming to fix anything.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -437,6 +441,39 @@ private Unsafe() {}
...
+    }
+
+    @Test public void invokeMethod_vs_DefaultGroovyMethods() throws Exception {
+        // Closure defines the invokeMethod method, and asBoolean is defined on DefaultGroovyMethods.
+        // the method dispatching in this case is that c.asBoolean() resolves to DefaultGroovyMethods.asBoolean()
+        // and not invokeMethod("asBoolean")
+
+        // calling asBoolean shouldn't go through invokeMethod
+        MetaMethod m1 = InvokerHelper.getMetaClass(ClassWithInvokeMethod.class).pickMethod("asBoolean",new Class[0]);
+        assert m1!=null;
```
amuniz: Why not use JUnit ```assert*``` methods here?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
jglick: Wrong, `invokeMethod_vs_DefaultGroovyMethods` is using this new entry from `GenericWhitelist`.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -437,6 +441,39 @@ private Unsafe() {}
...
+                        new GenericWhitelist(),
+                        new AbstractWhitelist() {
+                            @Override
+                            public boolean permitsConstructor(Constructor<?> constructor, Object[] args) {
+                                return constructor.getDeclaringClass()==ClassWithInvokeMethod.class;
+                            }
+                        }
+                ),
+                true,
+                "def c = new org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptorTest.ClassWithInvokeMethod(); c.asBoolean()"
```
jglick: Better to use `ClassWithInvokeMethod.class.getCanonicalName()`.

jglick:
:bee:

jglick:
:bee:

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
kohsuke: I should have split them in two commits or referenced that in the commit message, my bad.

I think we need to scan the `DefaultGroovyMethods` systematically and whitelist methods instead of waiting for bug reports to appear for each method one by one.

PR #20, opened 2015-08-04
[JENKINS-29541] GString receiver handling

jglick:
[JENKINS-29541](https://issues.jenkins-ci.org/browse/JENKINS-29541)

Adding a whitelist entry for `String.substring` is trivial and in fact an administrator could do so for themselves. The bigger problem was with `GString`, whose `invokeMethod` is a little hack to delegate unknown methods to `String`; this needs special treatment. (Ideally the Groovy sandbox library would handle this sort of painful detail, but it does not.)

@reviewbybees


amuniz:
:bee:

Out of scope: perhaps ```findMatchingMethod``` should be improved to search for the ```receiver``` method in parents hierarchy.

tfennelly:
:bee:

jglick:
> perhaps `findMatchingMethod` should be improved to search for the receiver method in parents hierarchy

Well that is what `method` does when it calls `findMatchingMethod` in a loop, but (so far as I know) `staticMethod` does not need that, nor does `method`’s check for `String` methods on a `GString`.

PR #22, opened 2015-08-18
Pick up a new version of groovy-sandbox that properly handles Map property access

kohsuke:
Also allow Map methods to be called from within sandbox.

Related to https://github.com/jenkinsci/workflow-plugin/pull/131

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -24,6 +24,12 @@ method java.lang.String substring int int
 method java.lang.String replace java.lang.CharSequence java.lang.CharSequence
 method java.util.Collection contains java.lang.Object
 method java.util.concurrent.Callable call
+method java.util.HashMap get java.lang.Object
+method java.util.HashMap put java.lang.Object java.lang.Object
+method java.util.LinkedHashMap get java.lang.Object
+method java.util.LinkedHashMap put java.lang.Object java.lang.Object
+method java.util.Map get java.lang.Object
+method java.util.Map put java.lang.Object java.lang.Object
 
 # Groovy:
 method groovy.lang.Script getBinding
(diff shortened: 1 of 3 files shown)


kohsuke:
@reviewbybees

kohsuke:
Tickling PRB again.

amuniz:
:bee:

jglick:
Could not find artifact org.kohsuke:groovy-sandbox:jar:1.10 in repo.jenkins-ci.org (http://repo.jenkins-ci.org/public/)

Did you forgot to close & release the staging repo?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -24,6 +24,12 @@ method java.lang.String substring int int
 method java.lang.String replace java.lang.CharSequence java.lang.CharSequence
 method java.util.Collection contains java.lang.Object
 method java.util.concurrent.Callable call
+method java.util.HashMap get java.lang.Object
+method java.util.HashMap put java.lang.Object java.lang.Object
+method java.util.LinkedHashMap get java.lang.Object
+method java.util.LinkedHashMap put java.lang.Object java.lang.Object
```
jglick: These should be redundant given the whitelists of `Map` methods, since the interceptor will look up the location of the method definition, not override/implementation. See for example

    method java.util.List get int

above.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -24,6 +24,12 @@ method java.lang.String substring int int
...
+method java.util.Map get java.lang.Object
+method java.util.Map put java.lang.Object java.lang.Object
```
jglick: This one was already there at line 12. I think only `get` was actually missing.

jglick:
Can you explain here what bug is being solved? If at all possible, it should be reproduced in `SandboxInterceptorTest`. (It has happened in the past that a problem can only be reproduced in conjunction with `groovy-cps` in `workflow-aggregator` tests, but this is unusual.)

jglick:
:bee: now.

daniel-beck:
:bee:

svanoort:
:bee:

kohsuke:
Thanks for adding a test! I should have done that.

PR #24, opened 2015-09-14
[JENKINS-30432] Warn about dangerous signatures

jglick:
In [JENKINS-30432](https://issues.jenkins-ci.org/browse/JENKINS-30432) for example, the user naïvely approved a signature that should never be approved. I have also seen @rsandell accidentally approve usage of `File` methods, which is even more dangerous. Introducing a blacklist of signatures known to be exploitable, or which would have no legitimate use, and sternly warning the administrator not to accept them.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/RejectedAccessException.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/RejectedAccessException.java
@@ -79,6 +79,7 @@ public final class RejectedAccessException extends SecurityException {
 
     /**
      * True if {@link #getSignature} is non-null but it would be a bad idea for an administrator to approve it.
+     * @since TODO 1.16?
      */
     public boolean isDangerous() {
         return dangerous;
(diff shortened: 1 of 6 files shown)


stephenc:
:bee:

oleg-nenashev:
:bee: 
I doubt it's possible to blacklist all known security flaws, but it's better than nothing

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/RejectedAccessException.java
```diff
@@ -76,4 +77,21 @@ public RejectedAccessException(String message) {
...
+    /**
+     * True if {@link #getSignature} is non-null but it would be a bad idea for an administrator to approve it.
+     */
+    public boolean isDangerous() {
+        return dangerous;
+    }
+
+    /**
+     * You may set this flag if you think it would be a security risk for this signature to be approved.
+     */
```
oleg-nenashev: :ant: - IllegalArgumentException should be documented IMO

oleg-nenashev:
re :bee:

PR #26, opened 2015-10-09
Ignore tests using old HtmlUnit methods

kwhetstone:
HtmlUnit was recently updated, causing certain tests that used methods in the old version of HtmlUnit to fail.  For now, we're going to skip these tests, test manually, and update the tests after the release of the next core.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
 
+import com.gargoylesoftware.htmlunit.ElementNotFoundException;
 import com.gargoylesoftware.htmlunit.html.HtmlCheckBoxInput;
 import com.gargoylesoftware.htmlunit.html.HtmlRadioButtonInput;
 import groovy.lang.Binding;
@@ -75,7 +76,7 @@ public class SecureGroovyScriptTest {
      * Basic approval test where the user doesn't have RUN_SCRIPTS privs but has unchecked
      * the sandbox checkbox. Should result in script going to pending approval.
      */
-    @Ignore("Upgrade issue; fix after next core update") @Test public void basicApproval() throws Exception {
+    @Test public void basicApproval() throws Exception {
         r.jenkins.setSecurityRealm(r.createDummySecurityRealm());
         GlobalMatrixAuthorizationStrategy gmas = new GlobalMatrixAuthorizationStrategy();
         gmas.add(Jenkins.READ, "devel");
@@ -88,7 +89,7 @@ public class SecureGroovyScriptTest {
         wc.login("devel");
         HtmlPage page = wc.getPage(p, "configure");
         HtmlForm config = page.getFormByName("config");
-        config.getButtonByCaption("Add post-build action").click(); // lib/hudson/project/config-publishers2.jelly
+        getButtonByCaption(config, "Add post-build action").click(); // lib/hudson/project/config-publishers2.jelly
         page.getAnchorByText(r.jenkins.getExtensionList(BuildStepDescriptor.class).get(TestGroovyRecorder.DescriptorImpl.class).getDisplayName()).click();
         HtmlTextArea script = config.getTextAreaByName("_.script");
         String groovy = "build.externalizableId";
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
```diff
@@ -75,7 +75,7 @@
      * Basic approval test where the user doesn't have RUN_SCRIPTS privs but has unchecked
      * the sandbox checkbox. Should result in script going to pending approval.
      */
-    @Test public void basicApproval() throws Exception {
+    @Ignore("Upgrade issue; fix after next core update") @Test public void basicApproval() throws Exception {
```
jglick: Not compilable.

jglick:
I would prefer to find some way to run these tests at least when using the old baseline. Ideally we could find some preferred methods that are present in both old and new versions of HtmlUnit—did they just deprecate and later delete the ones in use now? Or are there some lower-level, less convenient, but still-supported methods we could use?

At worst, wrap the relevant calls in

```java
try {
    someOldMethodSinceDeleted();
} catch (NoSuchMethodError x) {
    throw new AssumptionViolatedException(x);
}
```

jglick:
@tfennelly has [some advice](https://github.com/jenkinsci/jenkins/blob/master/test/src/main/java/com/gargoylesoftware/htmlunit/README.md#htmlelementutil-htmlformutil-and-domnodeutil) it seems.

jglick:
I suppose you meant this to be @reviewbybees but for now it is a :bug:

kwhetstone:
The method was likely deprecated and eventually removed; we're using a pretty old version of HtmlUnit in this plugin.  Since these extremely handy utility methods aren't in the stable-1.625, where this is failing, I added the missing method to this test.  Maybe we could merge the HtmlUtils changes in to 1.625?

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
oleg-nenashev: I would vote to print such messages on a separate line. I partially agree with Jesse's code style, but this long text pollutes the method signature IMO

jglick:
Currently uncompilable.

> Maybe we could merge the HtmlUtils changes in to 1.625?

Mm, perhaps. @tfennelly should advise.

(In the longer term I would like to split `jenkinsci/jenkins/test/src/main/` into its own repository with an independent lifecycle, so that plugins can pick a version of the test harness to use without needing to be tied to that shipped with their Jenkins core baseline.)

jglick:
:bug: for uncompilability (did not yet recheck diff)

tfennelly:
@kwhetstone From the changes, I can only see use of the old `getButtonByCaption` method. This is not available on the `HtmlFormUtil` class (as `getButtonByCaption`). Would probably be better to use that I think, but I don't see it as a bug if you do it yourself.

BTW ... `getButtonByCaption` was something KK added to the fork (as far as I can remember).

tfennelly:
Here's a link to the code: https://github.com/jenkinsci/jenkins/blob/master/test/src/main/java/com/gargoylesoftware/htmlunit/html/HtmlFormUtil.java#L127-L141

Looks quite like what you are already using.

jglick:
> Would probably be better to use that I think

Only possible when updating the core dep, unless you backport these utilities. See discussion above.

:bee:

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
```diff
@@ -842,4 +845,20 @@ public void testSandboxClassResolution() throws Exception {
...
         }
     }
+    
+    /**
+     * Get the form button having the specified text/caption.
+     * @param htmlForm The form containing the button.
+     * @param caption The button text/caption being searched for.
+     * @return The button if found.
+     * @throws ElementNotFoundException Failed to find the button on the form.
+     */
```
jglick: :ant: add a comment

```java
// TODO 1.627+ use standard version from HtmlFormUtil
```

so that we do not forget to clean up the temporary copy.

kwhetstone:
Added.  Thanks for the review.

tfennelly:
:bee: I think it's fine to use your own version of that util for now, adding an "update later" TODO.

I've read the comments, but still don't really get how you can be using a baseline/test-harness that doesn't use the updated `HtmlUnit`, yet isn't able to access the old methods.

stephenc:
:bee:

kwhetstone:
Thanks all; I'll come back and update the tests next release.

PR #32, opened 2015-11-09
Expand whitelist, specially with default groovy methods

ydubreuil:
@reviewbybees

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -7,7 +7,6 @@ method java.lang.CharSequence length
 method java.lang.Comparable compareTo java.lang.Object
 method java.lang.Object getClass
 method java.lang.Object equals java.lang.Object
-method java.lang.Object getClass
 method java.lang.Object toString
 staticMethod java.util.Arrays toString java.lang.Object[]
 method java.util.Collection add java.lang.Object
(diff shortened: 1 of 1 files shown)


rsandell:
:bee:

ydubreuil:
@reviewbybees done

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -7,6 +7,7 @@ method java.lang.CharSequence length
 method java.lang.Comparable compareTo java.lang.Object
 method java.lang.Object getClass
 method java.lang.Object equals java.lang.Object
+method java.lang.Object getClass
```
jglick: And what do you expect to do with the `Class` that would be safe? Generally anything smelling of reflection should be avoided unless it has been analyzed carefully.

jglick: …and anyway this entry was already there on line 8.

ydubreuil: @jglick I tried to be nasty with `Class` but failed to do anything bad. As you can see in the following stack trace, Sandbox is preventing use of reflexion:

```
org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use method java.lang.Class getField java.lang.String
	at org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.StaticWhitelist.rejectMethod(StaticWhitelist.java:150)
```

I will remove the duplicated line

jglick:
:bee:

PR #33, opened 2015-11-23
[JENKINS-31701] Misclassification of a method taking long and being passed an int

jglick:
[JENKINS-31701](https://issues.jenkins-ci.org/browse/JENKINS-31701)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
@@ -63,7 +63,7 @@ class GroovyCallSiteSelector {
             if (
                     parameterTypes[i].isPrimitive()
                     && parameters[i] != null
-                    && ClassUtils.primitiveToWrapper(parameterTypes[i]).isInstance(parameters[i])
+                    && isInstancePrimitive(ClassUtils.primitiveToWrapper(parameterTypes[i]), parameters[i])
             ) {
                 // Groovy passes primitive values as objects (for example, passes 0 as Integer(0))
                 // The prior test fails as int.class.isInstance(new Integer(0)) returns false.
@@ -78,6 +78,24 @@ class GroovyCallSiteSelector {
             return false;
         }
         return true;
+    }
+
+    private static boolean isInstancePrimitive(@Nonnull Class<?> type, @Nonnull Object instance) {
+        if (type.isInstance(instance)) {
+            return true;
+        }
+        // https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.1.2
+        if (instance instanceof Number) {
+            Number n = (Number) instance;
+            if (type == Long.class || type == Double.class) {
+                return true;
+            }
+            if (type == Integer.class && n.longValue() <= Integer.MAX_VALUE && n.longValue() >= Integer.MIN_VALUE) {
+                return true;
+            }
+            // TODO etc. for other types if they ever come up
+        }
+        return false;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -80,6 +80,24 @@ private static boolean matches(@Nonnull Class<?>[] parameterTypes, @Nonnull Obje
...
 
+    private static boolean isInstancePrimitive(@Nonnull Class<?> type, @Nonnull Object instance) {
+        if (type.isInstance(instance)) {
+            return true;
+        }
+        // https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.1.2
+        if (instance instanceof Number) {
+            Number n = (Number) instance;
+            if (type == Long.class || type == Double.class) {
+                return true;
```
kzantow: This looks like it would return true for possible narrowing conversions - that shouldn't happen, right?

jglick: `type` here is the method parameter type. If it is `long` then we can accept an `int` for `instance`. At line 93 we check whether a `long` could be narrowed to a method taking `int`.

kzantow: @jglick but a double could still be narrowed to a long in this case, yes?

jglick: Yes, the conversions between numeric types leave a lot of possibilities.

I just throw in `double` here although it is rarely used in interesting API methods. What I care about, and tested, is a method declared to take `long` but being passed an `int`, which caused @cyrille-leclerc pain. I am not interested in spending a lot of time fully implementing the JLS definitions unless there are known use cases, or there is some library out there which does the heavy lifting.

What we really want is for the Groovy runtime to select the actual `Method` (etc.) it is about to call, then pass that information to the `GroovyInterceptor` right before the call, so that we do not need to be endlessly tuning this plugin for special cases (in fact this whole class would disappear). Unfortunately the sandbox does not do that yet. See https://github.com/kohsuke/groovy-sandbox/issues/7.

oleg-nenashev:
@jglick 
I suppose <code>Byte</code> and <code>Short</code> types require the similar handling.

jglick:
> I suppose `Byte` and `Short` types require the similar handling.

And others besides; as already noted, there are plenty of corner cases not currently handled. But I do not have time to fix every case in the JLS; this is a fix for the known case that people run into. So I am looking for reviewers to confirm that this patch at least improves correctness and does not lead down a misguided path.

oleg-nenashev:
:bee: then

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -80,6 +80,24 @@ private static boolean matches(@Nonnull Class<?>[] parameterTypes, @Nonnull Obje
...
+            }
+            if (type == Integer.class && n.longValue() <= Integer.MAX_VALUE && n.longValue() >= Integer.MIN_VALUE) {
+                return true;
+            }
+            // TODO etc. for other types if they ever come up
+        }
+        return false;
+    }
```
tfennelly: I think this method could do with some javadoc/comments. For me, the name of the method (`isInstancePrimitive`) does not seem very accurate wrt what it seems like the method is actually doing (checking if `instance` can be widened/narrowed to `type`, right?). `isSafeToConvert(Object from, Class to)`?

tfennelly:
:bee: 

(but `isInstancePrimitive` seems like a bad name)

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
jglick: It is analogous to `Class.isInstance`, but tweaked to handle primitive values.

svanoort:
:bee:

PR #35, opened 2016-01-07
Clean up whitelists

jglick:
Validate content of whitelists and enforce a consistent sort order and uniqueness. Should make it easier to add new entries in parallel PRs while catching common mistakes.

@reviewbybees


andresrc:
:bee:

PR #36, opened 2016-01-14
[JENKINS-32493] Baseline 1.580.1 and Parent POM 2.3

andresrc:
* [x] Update baseline to 1.580.1
* [x] Update parent POM to 2.3
* [x] Adapt to JTH 2.1 and its updated HtmlUnit

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,10 +1,11 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
-    <parent>
-        <groupId>org.jenkins-ci.plugins</groupId>
-        <artifactId>plugin</artifactId>
-        <version>1.580.1</version>
-    </parent>
+  <parent>
+    <groupId>org.jenkins-ci.plugins</groupId>
+    <artifactId>plugin</artifactId>
+    <version>2.3</version>
+  </parent>
+
     <artifactId>script-security</artifactId>
     <version>1.18-SNAPSHOT</version>
     <packaging>hpi</packaging>
@@ -12,7 +13,8 @@
     <description>Allows Jenkins administrators to control what in-process scripts can be run by less-privileged users.</description>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Script+Security+Plugin</url> 
     <properties>
-      <findbugs-maven-plugin.version>3.0.2</findbugs-maven-plugin.version>
+      <jenkins.version>1.580.1</jenkins.version>
+      <java.level>6</java.level>
       <!--TODO: make true after the code cleanup-->
       <findbugs.failOnError>false</findbugs.failOnError>
     </properties>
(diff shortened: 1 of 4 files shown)


kwhetstone:
This change really simplifies the pom.  However, I notice that findbugs-maven-plugin.version is no longer configurable.  Is this something that's going to be missed with the new configuration?  I'm not sure if people are using this so it might be a non-issue.

andresrc:
@kwhetstone I don't have a strong opinion on that and would like to hear more. But my current position with regard to findbugs, and in the spirit of minimizing switches, is that we should aim for plugins using the "recommended" one. On the other side, I've seen many plugins configuring the possibility of not failing the build with findbugs errors, to handle transient cases. I think and including that switch in the parent pom, though not encouraging its use :smile:

stephenc:
@kwhetstone a property like `findbugs-maven-plugin.version` which is only used in one and only one place is an anti-pattern... it would be different if, for example, the maven plugin had the same release lifecycle as the annotations that it checks against and we therefore had to keep the annotations dependency in lock step with the maven plugin... in general:

* where there are two or more things that have to be kept in lock step, then you need to use a property
* where other functionality drives needing to configure from the CLI, then you need to use a property... the example is letting developers easily test their plugin against different versions of jenkins... `mvn -Djenkins.version=... hpi:run` in that case you may occasionally find that a newer version of jenkins has a requirement for a newer version of the hpi plugin... and more rarely a newer version of the stapler plugin... so in order to support the use case of letting you use `hpi:run` against different versions of jenkins we need to let the hpi plugin and the stapler plugin versions be controlled from the CLI.
* In all other cases, if the version is only defined in one place, *do not use a property*. Plugins that need to override the version can do so *the maven way* by overriding the version explicitly in `pluginManagement`

kwhetstone:
@stephenc thanks for the explanation.  I was primarily concerned about the case where you're using the cli since someone might have been using that variable for something.  I'm all for doing things the maven way.

pom.xml
```diff
@@ -12,7 +13,10 @@
     <description>Allows Jenkins administrators to control what in-process scripts can be run by less-privileged users.</description>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Script+Security+Plugin</url> 
     <properties>
-      <findbugs-maven-plugin.version>3.0.2</findbugs-maven-plugin.version>
+      <jenkins.version>1.509.4</jenkins.version>
+      <java.level>6</java.level>
+      <!-- TODO check if we'd better add the PI. -->
+      <jelly.requirePI>false</jelly.requirePI>
```
jglick: Yes you should.

pom.xml
```diff
@@ -28,19 +32,34 @@
...
+      <id>repo.jenkins-ci.org</id>
+      <url>http://repo.jenkins-ci.org/public/</url>
+      <releases>
+        <enabled>true</enabled>
+      </releases>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+    </pluginRepository>
+  </pluginRepositories>
```
jglick: What was wrong with the original versions of these sections?

pom.xml
```diff
@@ -55,8 +74,6 @@
         <dependency>
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>annotations</artifactId>
-            <version>3.0.0</version>
-            <scope>provided</scope>
         </dependency>
         <dependency> <!-- TODO default as of 1.568+ -->
```
jglick: Is this not now included by default in the parent POM? IMO it should be.

pom.xml
```diff
@@ -12,7 +13,10 @@
     <description>Allows Jenkins administrators to control what in-process scripts can be run by less-privileged users.</description>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Script+Security+Plugin</url> 
     <properties>
-      <findbugs-maven-plugin.version>3.0.2</findbugs-maven-plugin.version>
+      <jenkins.version>1.509.4</jenkins.version>
```
jglick: This is too old. I just checked and found that the oldest baseline used by a dependency of this plugin is now 1.580.1, so I might as well update to that here too. Will do so in a separate PR.

jglick: …#38

andresrc:
Addressed feedback and rebased.

pom.xml
```diff
@@ -12,7 +13,10 @@
     <description>Allows Jenkins administrators to control what in-process scripts can be run by less-privileged users.</description>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Script+Security+Plugin</url> 
     <properties>
-      <findbugs-maven-plugin.version>3.0.2</findbugs-maven-plugin.version>
+      <jenkins.version>1.580.1</jenkins.version>
+      <!-- TODO: Compilation errors to investigate when using 2.0 (matrix related) -->
+      <jenkins-test-harness.version>1.580.1</jenkins-test-harness.version>
```
jglick: :bug: probably you just need to add a test-scoped dependency on `matrix-auth`.

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest.java
```diff
@@ -179,4 +178,19 @@ public boolean handleConfirm(Page page, String message) {
...
     }
 
+    private void clickAndWait(HtmlElement e) throws IOException {
+        e.click();
+        e.getPage().getWebClient().waitForBackgroundJavaScript(10000);
+    }
+
+    private void click(HtmlPage page, String id, String value) throws IOException {
+        for (HtmlElement e : page.getElementById(id).getElementsByTagName("button")) {
+            if (e.hasAttribute("class") && value.equals(e.getAttribute("class"))) {
```
jglick: Yuck. Would it be easier to test if the script approval page just added some more `id`s?

andresrc: iirc it also tests it is removed from the pending list...

In any case the changes only tried to make it compatible with the upgraded HTMLUnit, not to refactor the test.

Thanks!

jglick:
> Update baseline to 1.580.1

is misleading, since that was already the baseline.

PR #40, opened 2016-01-21
Allow usage of isXXXX method as getter

Vlatombe:
@reviewbybees 

Related to https://github.com/jenkinsci/parallel-test-executor-plugin/pull/17

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
@@ -163,6 +163,14 @@ final class SandboxInterceptor extends GroovyInterceptor {
         if (m != null && whitelist.permitsMethod(m, receiver, args0)) {
             return super.onGetProperty(invoker, receiver, property);
         }
+        if (m == null) {
+            String booleanGetter = "is" + Functions.capitalize(property);
+            m = GroovyCallSiteSelector.method(receiver, booleanGetter, args0);
+            if (m != null && whitelist.permitsMethod(m, receiver, args0)) {
+                return super.onGetProperty(invoker, receiver, property);
+            }
+        }
+        System.out.println(m);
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -163,6 +163,14 @@
...
             return super.onGetProperty(invoker, receiver, property);
         }
+        if (m == null) {
+            String booleanGetter = "is" + Functions.capitalize(property);
+            m = GroovyCallSiteSelector.method(receiver, booleanGetter, args0);
+            if (m != null && whitelist.permitsMethod(m, receiver, args0)) {
+                return super.onGetProperty(invoker, receiver, property);
+            }
+        }
+        System.out.println(m);
```
jglick: delete please

jglick:
I think you forgot `static` methods; see (new) line 186.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -163,6 +163,14 @@
         if (m != null && whitelist.permitsMethod(m, receiver, args0)) {
             return super.onGetProperty(invoker, receiver, property);
         }
+        if (m == null) {
+            String booleanGetter = "is" + Functions.capitalize(property);
+            m = GroovyCallSiteSelector.method(receiver, booleanGetter, args0);
```
jglick: Unfortunately the way these methods work, you cannot “reuse” a local variable referring to candidate members without misreporting the `RejectedAccessException` in corner cases. Hence the long list of arguments to `rejectField`. I really need to refactor this stuff to make it easier to add clauses; probably need a helper method that takes a closure of some kind.

jglick:
:bug:, see #41.

PR #43, opened 2016-01-28
whitelist some more string functions

jtnord:
Whitelist some string regexp replacement functions and a char replacement function
@reviewbybees


jglick:
:bee:

PR #44, opened 2016-02-05
Add Entry getKey/getValue for java.util.Map

michaelneale:
It's nice to be able to get values when iterating over map entries (otherwise all you can do is get known key values out). 

@reviewbybees


rsandell:
:bee: would have liked to see some more map whitelisting while you were at it though ;)

andresrc:
:bee:

jglick:
:bee: but beware that this is not readily useful for Pipeline since typical `Map.Entry` implementations (e.g., `HashMap.Node`) are not `Serializable`. You would have to wrap any loops using entries in `@NonCPS`.

michaelneale:
@jglick yes I have had to be careful (however, working around serializable is generally less hassle than whitelist). 

@rsandell yes, I didn't want to be greedy. 

Thanks folks.

PR #45, opened 2016-03-23
Blacklist methods which start local processes

jglick:
Sometimes people will try to run, e.g.,

```groovy
def result = 'curl http://server/api'.execute().text
```

which is inherently unsafe.

@reviewbybees


amuniz:
:bee:

oleg-nenashev:
:bee:

andresrc:
:bee:

PR #46, opened 2016-03-23
[JENKINS-27725] More DefaultGroovyMethods handling

jglick:
[JENKINS-27725](https://issues.jenkins-ci.org/browse/JENKINS-27725) and some related issues, such as handling of virtual fields (which would prevent you from using `Process.text` even if #45 did not blacklist the creation of the `Process`).

@reviewbybees


andresrc:
:bee:

PR #48, opened 2016-03-28
[JENKINS-28277] Demonstrating that curry/ncurry work with the sandbox

jglick:
Companion to https://github.com/cloudbees/groovy-cps/pull/21.

@reviewbybees esp. @kohsuke


kohsuke:
:bee:

PR #49, opened 2016-03-30
MetaClass delegate handling

jglick:
Found that a trick used to bind synthetic methods to existing classes was being rejected by the sandbox. Here it is assumed that the closure itself is implicitly whitelisted; `Whitelist` does not have any mechanism for listing non-Java “methods” like this.

@reviewbybees esp. @kohsuke

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
@@ -24,12 +24,18 @@
 
 package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
 
+import groovy.lang.GroovyRuntimeException;
+import groovy.lang.MetaMethod;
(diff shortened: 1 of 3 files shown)


kohsuke:
IIUC, this allows any closure invocation, such as those obtained via method pointer operation. Is this safe? Looking at the test case, it seems like this case should be checked against `String.getAnswer()`  method.

:bee: so as not to block Jesse in case I got it all wrong.

jglick:
I could check whether it applies also to method pointers. I am not too concerned even if it does, since `DefaultGroovyMethod.getMetaClass` and `MetaClass.setProperty` will still be rejected, so you could not use that to mount an attack—a trusted source would have to have installed the metamethod to begin with. Or are you saying something else @kohsuke?

kohsuke:
I think I confused myself with earlier question about allowing any closure invocation, so I take that back.

andresrc:
:bee:

kohsuke:
I consider my previous :bee: standing, but here's another one because process!!

:bee:

PR #50, opened 2016-04-03
[JENKINS-26481] DefaultGroovyMethods logic error

jglick:
With this fix plus https://github.com/cloudbees/groovy-cps/pull/23, the test in https://github.com/jenkinsci/workflow-plugin/pull/372 still fails, but now for the well-understood JENKINS-27421, just as if the sandbox were disabled.

@reviewbybees esp. @kohsuke

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
@@ -78,8 +78,13 @@ final class SandboxInterceptor extends GroovyInterceptor {
             selfArgs[0] = receiver;
             System.arraycopy(args, 0, selfArgs, 1, args.length);
             for (Class<?> dgmClass : DGM_CLASSES) {
-                if (GroovyCallSiteSelector.staticMethod(dgmClass, method, selfArgs) != null) {
-                    return onStaticCall(invoker, dgmClass, method, selfArgs);
+                Method dgmMethod = GroovyCallSiteSelector.staticMethod(dgmClass, method, selfArgs);
+                if (dgmMethod != null) {
+                    if (whitelist.permitsStaticMethod(dgmMethod, selfArgs)) {
+                        return super.onMethodCall(invoker, receiver, method, args);
+                    } else {
+                        throw StaticWhitelist.rejectStaticMethod(dgmMethod);
+                    }
                 }
             }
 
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
looks good to me :bee:

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -78,8 +78,13 @@
             selfArgs[0] = receiver;
             System.arraycopy(args, 0, selfArgs, 1, args.length);
             for (Class<?> dgmClass : DGM_CLASSES) {
-                if (GroovyCallSiteSelector.staticMethod(dgmClass, method, selfArgs) != null) {
-                    return onStaticCall(invoker, dgmClass, method, selfArgs);
+                Method dgmMethod = GroovyCallSiteSelector.staticMethod(dgmClass, method, selfArgs);
+                if (dgmMethod != null) {
+                    if (whitelist.permitsStaticMethod(dgmMethod, selfArgs)) {
+                        return super.onMethodCall(invoker, receiver, method, args);
```
amuniz: So this is calling `method` as a regular instance method even being defined as `staticMethod`. Not sure if I understand.

jglick: Welcome to Groovyland. https://github.com/jenkinsci/workflow-plugin/pull/372 demonstrates what was wrong and what now works with this PR. The user invokes `each` on some `Object` (such as a `List`), passing a `Closure` as argument.

Normally Groovy translates this call into `static DefaultGroovyMethods.each(Object, Closure)`, like any other “GDK” methods which are injected into types that did not originally support the method. So we check the whitelist on that. Prior to this PR, we then delegate back to Groovy’s call site logic with the `DefaultGroovyMethods` replacement, which normally works.

But after `groovy-cps` hacks at the method dispatch tables so that such an `each` call now delegates to `CpsDefaultGroovyMethods.each(Object, Closure)`, when running in the sandbox we were still forwarding to the unpatched method.

So this fix means that we continue to check the same whitelist entry, but we now go back to Groovy’s own official lookup table if the call is approved, just like we do for all other kinds of method calls, so the patch to `CpsDefaultGroovyMethods` takes effect.

kohsuke:
:bee:
Given the question from @amuniz , I think it'd be prudent to leave in-code comment about what the intent behind this change is, but the change itself is good.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
amuniz: Thanks for the explanation!

amuniz:
:bee:

PR #52, opened 2016-04-08
[JENKINS-24399] Don't allow class directories any more.

andresrc:
See [JENKINS-24399](https://issues.jenkins-ci.org/browse/JENKINS-24399).

Proposed solution after the discussion in the issue is removing the possibility of using class folder as classpath entries.

Includes https://github.com/jenkinsci/script-security-plugin/pull/4

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
@@ -65,15 +65,46 @@ public final class ClasspathEntry extends AbstractDescribableImpl<ClasspathEntry
         }
     }
 
-    static String urlToPath(URL url) {
+    /** Returns {@code null} if another protocol or unable to perform the conversion. */
+    private static File urlToFile(@Nonnull URL url) {
         if (url.getProtocol().equals("file")) {
             try {
-                return new File(url.toURI()).getAbsolutePath();
+                return new File(url.toURI());
             } catch (URISyntaxException x) {
                 // ?
             }
         }
-        return url.toString();
+        return null;
+    }
+
+    static String urlToPath(URL url) {
+        final File file = urlToFile(url);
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
```diff
@@ -65,15 +65,46 @@ static URL pathToURL(String path) throws MalformedURLException {
...
+     * in case a different class loader is used by other {@link Language} implementation.
+     */
+    static boolean isClassDirectoryURL(@Nonnull URL url) {
+        final File file = urlToFile(url);
+        if (file != null && file.isDirectory()) {
+            return true;
+            // If the URL is a file but does not exist we fallback to default behaviour
+            // as non existence will be dealt with when trying to use it.
+        }
+        return url.toExternalForm().endsWith("/");
```
jglick: Hmm, careful. While this will correctly allow `file:/tmp/mylibs.jar`, it will incorrectly forbid `jar:file:/tmp/mylibs.jar!/`, which is actually the preferred form (IMO). Same for `jar:http://server/mylibs.jar!/`.

I would suggest

```java
String u = url.toExternalForm();
return u.endsWith("/") && !u.startsWith("jar:");
```

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntryTest.java
```diff
@@ -39,9 +46,23 @@
         }
         assertEquals("jar:file:/tmp/x.jar!/subjar.jar", "jar:file:/tmp/x.jar!/subjar.jar");
```
jglick: Seems I screwed up in #2; this is supposed to be `assertRoundTrip`. Could you fix it please?

jglick:
:bee:

jglick:
:bee:

PR #54, opened 2016-04-18
allow toList method

davidkarlsen:


davidkarlsen:
Hm - this did not work out - closing.

PR #55, opened 2016-04-19
allow toList methods

davidkarlsen:
Whitelist staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toList variants since they are very useful and harmless


olivergondza:
@jglick, parameter whitelists would be extremely helpful here. WDYT?

jglick:
> parameter whitelists would be extremely helpful here

You can already use `*` as a wildcard for the type of a single parameter. (Not yet for the name of a method, or for a variable number of parameters.) So

    staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toList *

would probably have sufficed. Anyway, I see nothing objectionable to having an explicit list.

PR #56, opened 2016-04-19
whitelist random UUID generation

davidkarlsen:


jglick:
Great, should allow us to delete [this](https://github.com/jenkinsci/workflow-aggregator-plugin/blob/a83d86638e635444d13aaf2a9d07f7d23fd82087/demo/JENKINS_HOME/scriptApproval.xml#L8)!

PR #57, opened 2016-04-19
whitelist staticMethod java.lang.System currentTimeMillis

davidkarlsen:


jglick:
BTW feel free to batch these kinds of things up into one PR.

PR #58, opened 2016-04-21
Support enum initializers

magnusr:


magnusr:
Solves https://issues.jenkins-ci.org/browse/JENKINS-33023

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -140,10 +140,28 @@ private static boolean isInstancePrimitive(@Nonnull Class<?> type, @Nonnull Obje
         return findMatchingMethod(receiver, method, args);
     }
 
+    private static boolean isEnumInitializer(Method m) {
+        if (!m.getDeclaringClass().isEnum()) {
+            return false;
+        }
+        if (!m.getName().equals("$INIT")) {
```
jglick: Hmm, is this name guaranteed by the JLS? Not a real problem if it is not, just wondering if we can make this a bit more portable.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -556,6 +556,20 @@ public Object invokeMethod(String name, Object args) {
...
+        String script = "enum Thing {\n"
+            + "  FIRST(\"The first thing\");\n"
+            + "  String description;\n"
+            + "  public Thing(String description) {\n"
+            + "    this.description = description;\n"
+            + "  }\n"
+            + "}\n"
+            + "Thing.values()[0].description\n";
+        String expected = "The first thing";
+        assertEvaluate(new BlanketWhitelist(), expected, script);
```
jglick: Best not to use `BlanketWhitelist` in tests like this. Rather you want to use either `GenericWhitelist`, or an explicit `StaticWhitelist`.

Anyway although this PR looks like it will fix the `unclassified staticMethod` error, it does not go far enough. We do not want to force the user to whitelist anything we know is safe. In the case of the script shown above, I would have expected [this code](https://github.com/jenkinsci/script-security-plugin/blob/1c321f31b5b382390d2509a699cd7516f7e25a5c/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java#L136) to automatically whitelist the initializer.

Perhaps this is just a matter of the test not reflecting what the code change is actually accomplishing.

jglick: cf. `whitelistedIrrelevantInsideScript`

jglick: Also I would expect `.values()` to be automatically whitelisted; the current test does not indicate whether it is or not.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
magnusr: It appears it isn't. How about checking that the method is synthetic instead? No other method on an enum type should take `Object[]` as arguments. That said, I don't think the JLS guarantees that the argument types will be `Object[]`, so it's still not perfect, only better.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
magnusr: Yup, as you showed in your follow-up, GenericWhitelist was enough. The BlanketWhitelist was a copy-paste mistake on my behalf.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
jglick: > How about checking that the method is synthetic instead?

Might work. Try it and file a follow-up if it works.

magnusr: https://github.com/jenkinsci/script-security-plugin/pull/65

jglick:
FTR the synthetic method is a varargs

```java
public static final $INIT(Object...)
```

which gets called with the `name` and `ordinal` of the enum value, followed by the constructor arguments. The special handling in `isEnumInitializer` could perhaps be replaced by a more general check for a `isVarArgs` method that could match the supplied arguments; I am not sure if the interceptor normally receives the `Object[]` in such cases.

Enums defined in Java code seem to use different bytecode. These constructors are not caught by the interceptor at all, since to Groovy the resulting values just look like `static` fields.

PR #60, opened 2016-04-22
Blacklist new java.io.FileInputStream java.lang.String

jglick:
Saw someone trying to use this from a `Jenkinsfile`. It is dangerous, and usually better done with `readFile`.

The constructor overload taking `File` is not a thread because we are already blacklisting the `File` constructors.

@reviewbybees


recena:
:bee: since It seems a bad practice

escoem:
a quick 🐝  here

PR #61, opened 2016-04-26
[JENKINS-33023] Making sure we can really use enums without extra whitelisting

jglick:
#58 follow-up.

@reviewbybees and @magnusr


andresrc:
:bee:

magnusr:
Yup, looks like what I should have done in the first place =) Thanks!

kwhetstone:
:bee:

PR #53, opened 2016-04-14
[JENKINS-34251] Some additional tests

andresrc:
See [JENKINS-34251](https://issues.jenkins-ci.org/browse/JENKINS-34251)

Small additions so that for both scripts and classpath entries we have:
* Configure and use with and without security.
* Full cycle of several entities checking from the model and from the UI.

Most of the changes are to factor out the common parts.

@reviewbybees

PR #63, opened 2016-04-27
[JENKINS-34466] Fix PCT against 2.0

andresrc:
When building against Jenkins 2.0 Groovy 2.4.6 is used instead of 1.8.9 in the baseline. This brings a couple of issues:
* For some methods in the whitelist there are additional overloads for more specific type arguments, generating approval errors.
* The behavior for ambiguous overloading has changed: no longer "some" version of the method is picked but an exception is thrown.

Besides, the PR includes:
* Findbugs fixes when building against 2.0
* Some cleanup after https://github.com/jenkinsci/script-security-plugin/pull/53

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelist.java
@@ -43,16 +43,17 @@ import java.util.List;
 import java.util.Set;
 import javax.annotation.Nonnull;
 
+import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import groovy.lang.GroovySystem;
+import hudson.util.VersionNumber;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException;
 
 /**
  * Whitelist based on a static file.
  */
 public final class StaticWhitelist extends EnumeratingWhitelist {
-    private static final String G2_PREFIX = "g2";
-    private static final boolean GROOVY2 = GroovySystem.getVersion().startsWith("2.");
+    static final boolean GROOVY2 = new VersionNumber(GroovySystem.getVersion()).compareTo(new VersionNumber("2")) >= 0;
 
     final List<MethodSignature> methodSignatures = new ArrayList<MethodSignature>();
     final List<NewSignature> newSignatures = new ArrayList<NewSignature>();
@@ -86,21 +87,10 @@ public final class StaticWhitelist extends EnumeratingWhitelist {
      * @param line Line to filter.
      * @return {@code null} if the like must be skipped or the content to process if not.
      */
-    static String filter(String line) {
-        if (line == null) {
-            return null;
-        }
+    static @CheckForNull String filter(@Nonnull String line) {
         line = line.trim();
         if (line.isEmpty() || line.startsWith("#")) {
             return null;
-        }
-        if (line.startsWith(G2_PREFIX)) {
-            if (GROOVY2) {
-                return line.substring(G2_PREFIX.length()).trim();
-                // If empty or malformed after the prefix, we'll leave the parsing fail.
-            } else {
-                return null; // skip
-            }
         }
         return line;
     }
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelist.java
@@ -44,12 +44,15 @@ import java.util.Set;
 import javax.annotation.Nonnull;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import groovy.lang.GroovySystem;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException;
 
 /**
  * Whitelist based on a static file.
  */
 public final class StaticWhitelist extends EnumeratingWhitelist {
+    private static final String G2_PREFIX = "g2";
+    private static final boolean GROOVY2 = GroovySystem.getVersion().startsWith("2.");
 
     final List<MethodSignature> methodSignatures = new ArrayList<MethodSignature>();
     final List<NewSignature> newSignatures = new ArrayList<NewSignature>();
@@ -61,11 +64,10 @@ public final class StaticWhitelist extends EnumeratingWhitelist {
         BufferedReader br = new BufferedReader(definition);
         String line;
         while ((line = br.readLine()) != null) {
-            line = line.trim();
-            if (line.isEmpty() || line.startsWith("#")) {
-                continue;
+            line = filter(line);
+            if (line != null) {
+                add(line);
             }
-            add(line);
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -127,7 +127,14 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods asBoolean java.lan
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods collect java.lang.Object groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods collect java.util.Collection groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods disjoint java.util.Collection java.util.Collection
+g2 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods each java.lang.Iterable groovy.lang.Closure
```
andresrc: New overloads in Groovy 2.4.6 not in 1.8.9

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/AbstractApprovalTest.java
```diff
@@ -39,13 +39,9 @@
...
-        configureSecurity();
-        create().assertCannotUse();
-    }
-
-    @Test public void withSecurityAccepted() throws Exception {
+    @Test public void withSecurity() throws Exception {
         configureSecurity();
+        // Cannot use until approved
         create().assertCannotUse().approve().use();
     }
```
andresrc: Cleanup from https://github.com/jenkinsci/script-security-plugin/pull/53

armfergom:
This LGTM as far as I understand. 🐝

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelist.java
```diff
@@ -44,12 +44,15 @@
...
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import groovy.lang.GroovySystem;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException;
 
 /**
  * Whitelist based on a static file.
  */
 public final class StaticWhitelist extends EnumeratingWhitelist {
+    private static final String G2_PREFIX = "g2";
+    private static final boolean GROOVY2 = GroovySystem.getVersion().startsWith("2.");
```
jglick: :ant: I think you want to use `VersionNumber` or similar to apply the same behavior (unless and until overridden) in Groovy 3.x.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelist.java
```diff
@@ -79,6 +81,30 @@ public StaticWhitelist(String... lines) throws IOException {
         this(asList(lines));
     }
 
+    /**
+     * Filters a line, returning the content that must be processed.
+     * @param line Line to filter.
+     * @return {@code null} if the like must be skipped or the content to process if not.
+     */
+    static String filter(String line) {
```
jglick: use `@CheckForNull`

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelist.java
```diff
@@ -79,6 +81,30 @@ public StaticWhitelist(String... lines) throws IOException {
...
+        if (line == null) {
+            return null;
```
jglick: :ant: AFAICT both callers always pass non-null, so delete this and make parameter `@Nonnull`

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
jglick: :bug: this is not maintainable, and there are/will be other libraries with different versions (including Jenkins core). I suppose the only reason not to just add the signature with no prefix is so that `GenericWhitelistTest.sanity` does not complain? Then just add a separate `generic-whitelist-groovy2` file and do not include it in `sanity` (or include it only if we can detect that the Groovy version in the classpath is sufficiently new).

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -492,10 +495,28 @@ private Unsafe() {}
...
+    @Test public void ambiguousOverloads18() {
+        if (GroovySystem.getVersion().startsWith("1.")) {
+            // In 1.8.9 Groovy selects one of these. How, I do not know.
+            assertAmbiguity("null");
+        } else {
+            throw new AssumptionViolatedException("Not using Groovy 1.x");
+        }
+    }
+
+    @Test(expected = GroovyRuntimeException.class) public void ambiguousOverloads24() {
```
jglick: :ant: rather use `assertEvaluate` with a script catching `GroovyRuntimeException`

jglick: (meaning inline `assertAmbiguity` back into `ambiguousOverloads18`)

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
jglick: (treating as a bug not an ant because this is a format change and thus not easily reverted)

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -756,14 +757,24 @@ private synchronized void removePendingScript(String hash) {
         return pendingSignatures;
     }
 
+    // Added to account for new findbugs annotations in ExtensionList#get (PCT scenario)
```
jglick: Under PCT, compilation and FindBugs would still be done against the old baseline, so why is this here?

andresrc: For these plugins, besides the current scenario, we are also doing full builds against both baseline and target.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelist.java
```diff
@@ -43,13 +43,17 @@
...
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import groovy.lang.GroovySystem;
+import hudson.util.VersionNumber;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException;
 
 /**
  * Whitelist based on a static file.
  */
 public final class StaticWhitelist extends EnumeratingWhitelist {
+    static final boolean GROOVY2 = new VersionNumber(GroovySystem.getVersion()).compareTo(new VersionNumber("2")) >= 0;
```
jglick: :ant: inline into `GenericWhitelistTest`, its only remaining usage.

jglick:
:bee:

PR #67, opened 2016-05-06
[JENKINS-34741] Reproduced problem in test

jglick:
[JENKINS-34741](https://issues.jenkins-ci.org/browse/JENKINS-34741)

@oleg-nenashev passed on a report of this. Not yet obvious to me how to fix, but useful to have the test case recorded.

@reviewbybees


abayer:
:bee: Yeah, short of whitelisting named parameter constructors in a blanket way, I'm not sure what the solution would be.

jglick:
According to [this](http://groovy-lang.org/style-guide.html#_initializing_beans_with_named_parameters_and_the_default_constructor) the sandbox should be checking whitelists for the default constructor and then all mentioned setters (or I guess field accesses).

Yet another case of the sandbox not telling us what it is really running; cf. https://github.com/kohsuke/groovy-sandbox/issues/7.

oleg-nenashev:
:bee:, but we need to create a JIRA issues and reference it before merging

PR #68, opened 2016-05-11
[JENKINS-34739] Handle varargs

jglick:
[JENKINS-34739](https://issues.jenkins-ci.org/browse/JENKINS-34739)

Replaces the fix in #58 (and the follow-up in #65).

@reviewbybees and CC @magnusr

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
@@ -26,6 +26,7 @@ package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
 
 import com.google.common.primitives.Primitives;
 import groovy.lang.GString;
+import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
@@ -43,7 +44,27 @@ import org.apache.commons.lang.ClassUtils;
  */
 class GroovyCallSiteSelector {
 
-    private static boolean matches(@Nonnull Class<?>[] parameterTypes, @Nonnull Object[] parameters) {
+    private static boolean matches(@Nonnull Class<?>[] parameterTypes, @Nonnull Object[] parameters, boolean varargs) {
+        if (varargs) {
+            int fixedLen = parameterTypes.length - 1;
+            Class<?> componentType = parameterTypes[fixedLen].getComponentType();
+            assert componentType != null;
+            int arrayLength = parameters.length - fixedLen;
+            if (arrayLength >= 0) {
+                if (arrayLength == 1 && parameterTypes[fixedLen].isInstance(parameters[fixedLen])) {
+                    // not a varargs call
+                } else {
+                    Object array = Array.newInstance(componentType, arrayLength);
+                    for (int i = 0; i < arrayLength; i++) {
+                        Array.set(array, i, parameters[fixedLen + i]);
+                    }
+                    Object[] parameters2 = new Object[fixedLen + 1];
+                    System.arraycopy(parameters, 0, parameters2, 0, fixedLen);
+                    parameters2[fixedLen] = array;
+                    parameters = parameters2;
+                }
+            }
+        }
         if (parameters.length != parameterTypes.length) {
             return false;
         }
@@ -128,7 +149,7 @@ class GroovyCallSiteSelector {
 
     public static @CheckForNull Constructor<?> constructor(@Nonnull Class<?> receiver, @Nonnull Object[] args) {
         for (Constructor<?> c : receiver.getDeclaredConstructors()) {
-            if (matches(c.getParameterTypes(), args)) {
+            if (matches(c.getParameterTypes(), args, c.isVarArgs())) {
                 return c;
             }
         }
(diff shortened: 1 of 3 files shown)


magnusr:
👍

kwhetstone:
:bee:

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -43,7 +44,27 @@
...
-    private static boolean matches(@Nonnull Class<?>[] parameterTypes, @Nonnull Object[] parameters) {
+    private static boolean matches(@Nonnull Class<?>[] parameterTypes, @Nonnull Object[] parameters, boolean varargs) {
+        if (varargs) {
+            int fixedLen = parameterTypes.length - 1;
+            Class<?> componentType = parameterTypes[fixedLen].getComponentType();
+            assert componentType != null;
+            int arrayLength = parameters.length - fixedLen;
+            if (arrayLength >= 0) {
+                if (arrayLength == 1 && parameterTypes[fixedLen].isInstance(parameters[fixedLen])) {
+                    // not a varargs call
```
svanoort: 🐛  Rather than redefining an argument in a fairly complex code block, let's move this to a separate method returning the new parameters (hard to follow otherwise).

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -43,7 +44,27 @@
...
+                } else {
+                    Object array = Array.newInstance(componentType, arrayLength);
+                    for (int i = 0; i < arrayLength; i++) {
+                        Array.set(array, i, parameters[fixedLen + i]);
```
svanoort: I assume we're doing the set operation rather than arraycopy to handle primitives and boxing/unboxing?

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -140,28 +161,10 @@ private static boolean isInstancePrimitive(@Nonnull Class<?> type, @Nonnull Obje
         return findMatchingMethod(receiver, method, args);
     }
 
-    private static boolean isEnumInitializer(Method m) {
```
svanoort: Er, why did we remove this?

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
jglick: No, could probably have used `arraycopy`.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
jglick: Because is has been superseded by the more general fix.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
svanoort: 🐜 then I think this could be done more simply with one of the Arrays.copyOfRange methods?  Ditto the other one too, I think, reducing this to three simple lines.

jglick: Never mind, you _do_ get an `ArrayStoreException`. Needs test coverage.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
svanoort: To capture reasoning discussed OoB for making this a bug and not a nit:  

> Redefining inputs introduces unexpected behavior that others may not catch without reading the function in entirety and understanding it deeply. 

Also 🐜  that we can remove the initial if case, since it's a no-op.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
jglick: > then I think this could be done more simply with one of the `Arrays.copyOfRange` methods?

No, these are for arrays whose component type is statically known.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -102,6 +85,33 @@ private static boolean matches(@Nonnull Class<?>[] parameterTypes, @Nonnull Obje
...
+            if (arrayLength == 1 && parameterTypes[fixedLen].isInstance(parameters[fixedLen])) {
+                // not a varargs call
+                return parameters;
+            } else {
+                Object array = Array.newInstance(componentType, arrayLength);
+                for (int i = 0; i < arrayLength; i++) {
+                    Array.set(array, i, parameters[fixedLen + i]);
+                }
+                Object[] parameters2 = new Object[fixedLen + 1];
+                System.arraycopy(parameters, 0, parameters2, 0, fixedLen);
```
svanoort: Super nitpicky 🐜 - can create and copy parameters2 with a single copyOfRange call (null will be placed in fixedLen until set).

svanoort:
🐝 - the method refactor makes it isn't doing somewhat unexpected things with inputs.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -471,6 +475,14 @@ public static int len(String... vals) {
             return vals.length;
         }
         @Whitelisted
+        public static int sum(int... numbers) {
```
svanoort: 😄  for seeing this tested.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
jglick: Yeah. Could.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
svanoort: Object is a type... though the primitives issues break this so no point now.

recena:
:bee:

PR #71, opened 2016-05-20
Groovy 2.x seems to define .each() method for different receiver type…

kohsuke:
…s separately.

TODO: whitelist that allows all overloaded versions


jglick:
```
java.lang.AssertionError: staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods each java.util.List groovy.lang.Closure does not exist (or is an override)
	at org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.GenericWhitelistTest.sanity(GenericWhitelistTest.java:32)
```

needs to be moved to the 2.x list.

jglick:
Superseded by #63.

PR #72, opened 2016-05-27
[FIXED JENKINS-35191] Add Map.keySet() and Map.value() to whitelist.

abayer:
https://issues.jenkins-ci.org/browse/JENKINS-35191

cc @reviewbybees @jglick


jglick:
:bee:

PR #80, opened 2016-08-12
[JENKINS-23784] Cache upstream class loader results

jglick:
Works around [JENKINS-23784](https://issues.jenkins-ci.org/browse/JENKINS-23784) as well as the general expense of calling into core/plugin class loaders repeatedly during Groovy compilation.

@reviewbybees


abayer:
:bee: lgtm

PR #81, opened 2016-08-19
Adding some more file-related blacklist entries

jglick:
Stuff I found in a user’s script that should not have been there.

@reviewbybees


rsandell:
:bee:

PR #84, opened 2016-09-21
[JENKINS-37599] Prevent empty classpath entries from being saved accidentally

jglick:
[JENKINS-37599](https://issues.jenkins-ci.org/browse/JENKINS-37599)

At least makes the problem clearer.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
@@ -68,7 +68,7 @@ public final class ClasspathEntry extends AbstractDescribableImpl<ClasspathEntry
             if (f.isAbsolute()) {
                 return f.toURI().toURL();
             } else {
-                throw new MalformedURLException("classpath entries must be URLs or absolute file paths");
+                throw new MalformedURLException("Classpath entry ‘" + path + "’ does not look like either a URL or an absolute file path");
             }
         }
     }
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
@@ -58,10 +58,18 @@ public final class ClasspathEntry extends AbstractDescribableImpl<ClasspathEntry
     }
     
     static URL pathToURL(String path) throws MalformedURLException {
+        if (path.isEmpty()) {
+            throw new MalformedURLException("JENKINS-37599: empty classpath entries not allowed");
+        }
         try {
             return new URL(path);
         } catch (MalformedURLException x) {
-            return new File(path).toURI().toURL();
+            File f = new File(path);
+            if (f.isAbsolute()) {
+                return f.toURI().toURL();
+            } else {
+                throw new MalformedURLException("classpath entries must be URLs or absolute file paths");
+            }
         }
     }
 
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
@@ -28,7 +28,6 @@ import com.gargoylesoftware.htmlunit.ElementNotFoundException;
 import com.gargoylesoftware.htmlunit.html.HtmlButton;
 import com.gargoylesoftware.htmlunit.html.HtmlCheckBoxInput;
 import com.gargoylesoftware.htmlunit.html.HtmlElement;
-import com.gargoylesoftware.htmlunit.html.HtmlRadioButtonInput;
 import groovy.lang.Binding;
 import hudson.remoting.Which;
 import org.apache.tools.ant.AntClassLoader;
@@ -46,7 +45,6 @@ import hudson.security.Permission;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.Publisher;
 import java.io.File;
-import java.io.FileFilter;
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
@@ -57,7 +55,6 @@ import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.tools.ant.DirectoryScanner;
 import org.apache.tools.ant.taskdefs.Expand;
-import org.apache.tools.ant.taskdefs.Touch;
 import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
 import org.jenkinsci.plugins.scriptsecurity.scripts.UnapprovedUsageException;
 import static org.junit.Assert.*;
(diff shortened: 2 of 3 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
```diff
@@ -58,10 +58,18 @@ public ClasspathEntry(@Nonnull String path) throws MalformedURLException {
...
+            throw new MalformedURLException("JENKINS-37599: empty classpath entries not allowed");
+        }
         try {
             return new URL(path);
         } catch (MalformedURLException x) {
-            return new File(path).toURI().toURL();
+            File f = new File(path);
+            if (f.isAbsolute()) {
+                return f.toURI().toURL();
+            } else {
```
jglick: `""` was being misinterpreted as a file path. Handling that directly now, but also the more general case of bogus text.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
```diff
@@ -287,20 +284,16 @@
...
-        
-        JenkinsRule.WebClient wc = r.createWebClient();
-        r.submit(wc.getPage(p, "configure").getFormByName("config"));
-        
-        p = r.jenkins.getItemByFullName(p.getFullName(), FreeStyleProject.class);
-        TestGroovyRecorder recorder = (TestGroovyRecorder)p.getPublishersList().get(0);
-        assertEquals(classpath, recorder.getScript().getClasspath());
+        ));
+        TestGroovyRecorder recorder2 = r.configRoundtrip(recorder);
+        r.assertEqualDataBoundBeans(recorder, recorder2);
```
jglick: Just simplifying original test from @ikedam in #2.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
```diff
@@ -58,10 +58,18 @@ public ClasspathEntry(@Nonnull String path) throws MalformedURLException {
     }
     
     static URL pathToURL(String path) throws MalformedURLException {
+        if (path.isEmpty()) {
```
oleg-nenashev: StringUtils#isBlank() ?

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
```diff
@@ -287,20 +284,16 @@
...
+        classpath.clear();
+        recorder2 = r.configRoundtrip(recorder);
+        r.assertEqualDataBoundBeans(recorder, recorder2);
```
jglick: This just verifies that the normal case of empty classpaths works in the normal usage mode from, say, Groovy Postbuild. Does not attempt to reproduce bug in Promoted Builds, which would require (for example) a separate mock publisher that has an initially nonempty `List<SecureGroovyScript>`.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
```diff
@@ -58,10 +58,18 @@ public ClasspathEntry(@Nonnull String path) throws MalformedURLException {
...
+        }
         try {
             return new URL(path);
         } catch (MalformedURLException x) {
-            return new File(path).toURI().toURL();
+            File f = new File(path);
+            if (f.isAbsolute()) {
+                return f.toURI().toURL();
+            } else {
+                throw new MalformedURLException("classpath entries must be URLs or absolute file paths");
```
oleg-nenashev: Add the specified path to the exception message. :ant:

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
jglick: Well, spaces would imply that the user actually deliberately typed something in the textbox, which would be a different case.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
svanoort: Second 🐜

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
svanoort: The isAbsolute test should catch that right?

aheritier:
LGTM but I'll wait the build success to give a bee

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
svanoort: 🐜  though I'm not 100% sold on isAbsolute & URL conversion as the validation.  Feels like it might have holes there.

svanoort:
Holding back on my bee until comments are addressed but will give it my vote if my concerns about edge cases slipping through are addressed.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
oleg-nenashev: Likely yes

oleg-nenashev:
🐝

oleg-nenashev:
maybe it worts to perform null path check btw. Just in case

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
jglick: see `ClasspathEntryTest.pathToURL`

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
jglick: @svanoort this was already addressed, look at the revised diff please (not sure what happened to “commented on an ignored diff” in @github)

jglick:
> null path check

Stapler will not pass nulls.

jglick:
> Stapler will not pass nulls.

Well, this was a bit of a misstatement. Stapler will just process JSON and pass whatever fields there are in JSON. But the JSON will always contain the `path` field, since there is a textbox, and the browser will submit it whether there is text in it or not.

Anyway calling `new URL(null)` will result in a `MalformedURLException` so there is no risk.

svanoort:
🐝

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
svanoort: I meant the spaces, but yeah I see your new test and LGTM

aheritier:
🐝

PR #86, opened 2016-09-23
Permit access to the messages of exceptions.

dothebart:
Currently Scripts may catch exceptions, but not fetch their messages for i.e. putting it into error report mails.


dothebart:
ok, I was trying to find a sensitive place. is it alphabetic order?

dothebart:
it seems that jenkins mixes commits to pr #85 into the tests it does? seems like a bug in jenkins to me?

jglick:
No, your commit just broke an existing test.

dothebart:
Ah, Ic. thanks for the assistance.

PR #39, opened 2016-01-20
[JENKINS-33024] whitelist XmlSlurper and associated features

cyrille-leclerc:
Whitelist XmlSlurper and associated features + whitelist the StringWriter API

:warning:  I had to whitelist  `method groovy.lang.GroovyObject getProperty java.lang.String` 
@jglick can you confirm that it's safe?

PR #87, opened 2016-09-23
Update generic-whitelist

rsandell:
Some hand picked white listings from our production instance that I believe are OK.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -35,6 +35,15 @@ method groovy.lang.Closure setDelegate java.lang.Object
 method groovy.lang.Closure setResolveStrategy int
 method groovy.lang.GString plus java.lang.String
 method groovy.lang.Script getBinding
+method groovy.util.Node get java.lang.String
+method groovy.util.NodeList getAt java.lang.String
+
+#There are other issues whith these classes than the whitelisting, but should at least be here
+new groovy.util.XmlParser
+new groovy.util.XmlSlurper
+method groovy.util.XmlParser parseText java.lang.String
+method groovy.util.XmlSlurper parseText java.lang.String
+new groovy.xml.MarkupBuilder java.io.Writer
 
 # Needed for Groovy Templates to emit output:
 method java.io.PrintWriter print java.lang.Object
@@ -42,11 +51,18 @@ method java.io.PrintWriter print java.lang.String
 
 # Useful general stuff:
 new java.io.StringReader java.lang.String
+new java.io.StringWriter
+new java.lang.Boolean java.lang.String
+staticMethod java.lang.Boolean parseBoolean java.lang.String
+staticMethod java.lang.Boolean valueOf boolean
+staticMethod java.lang.Boolean valueOf java.lang.String
 method java.lang.CharSequence charAt int
 method java.lang.CharSequence length
+field java.lang.Class name
 method java.lang.Comparable compareTo java.lang.Object
 method java.lang.Enum name
 method java.lang.Enum ordinal
+method java.lang.Iterable iterator
 # could add valueOf, though currently the staticField’s need to be whitelisted, which is the more likely use case
 method java.lang.Number byteValue
 method java.lang.Number doubleValue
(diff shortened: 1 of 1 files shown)


rsandell:
@reviewbybees esp. @jglick

jglick:
`GenericWhitelistTest.sanity` checks that things are listed in order, please update accordingly.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -35,18 +35,34 @@ method groovy.lang.Closure setDelegate java.lang.Object
...
 method groovy.lang.Script getBinding
+method groovy.util.Node get java.lang.String
+method groovy.util.NodeList getAt java.lang.String
+
+#There are other issues whith these classes than the whitelisting, but should at least be here
+new groovy.util.XmlParser
+new groovy.util.XmlSlurper
+method groovy.util.XmlParser parseText java.lang.String
+method groovy.util.XmlSlurper parseText java.lang.String
+new groovy.xml.MarkupBuilder java.io.Writer
```
jglick: See #39; I would rather this be pulled out and considered separately with end-to-end tests.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -35,18 +35,34 @@ method groovy.lang.Closure setDelegate java.lang.Object
...
 # Useful general stuff:
 new java.io.StringReader java.lang.String
+new java.io.StringWriter
+new java.lang.Boolean java.lang.String
+staticMethod java.lang.Boolean parseBoolean java.lang.String
+staticMethod java.lang.Boolean valueOf boolean
+staticMethod java.lang.Boolean valueOf java.lang.String
 method java.lang.CharSequence charAt int
 method java.lang.CharSequence length
+field java.lang.Class name
```
jglick: I think you wanted

    method java.lang.Class getName

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -128,19 +155,58 @@ staticField java.util.Locale SIMPLIFIED_CHINESE
...
 method java.util.Map$Entry getValue
 staticMethod java.util.TimeZone getTimeZone java.lang.String
+new java.util.TreeSet
 staticMethod java.util.UUID randomUUID
 method java.util.concurrent.Callable call
+method java.util.regex.MatchResult group int
+method java.util.regex.Matcher matches
 method java.util.regex.Matcher replaceAll java.lang.String
 method java.util.regex.Matcher replaceFirst java.lang.String
+method javax.xml.parsers.DocumentBuilder parse org.xml.sax.InputSource
```
jglick: This might be dangerous. IIRC it allows any URL or whatever.

I think it would be better to encapsulate things like XML or JSON parsing into the Pipeline Utility Steps plugin, where you could limit it to a `String` input (for example from `readFile`).

rsandell: Well this white list is for  InputSource and iirc there is no `new InputSourceX` white listed that would be dangerous?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
rsandell: I think this was for `whatever.class.name` hence the field instead of `getName()`

rsandell:
I added the  "xml stuff" because if the issues I've had with `XmlSlurper` et. al. and most of the XML manipulations we had to do is not simple to add into a generic utility step without replicating the entire `javax.xml` package. ;)

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
jglick: But whitelisting `getName()` means that Groovy will call `getName()` when you write `.name`, which is what we want.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
jglick: Perhaps so…would need to take a close look at all of these to be sure.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
rsandell: ok

jtnord:
@rsandell any update - we especially want things like `Boolean.valueOf(String)`

jglick:
Delete all XML-related APIs and I will reconsider this.

PR #93, opened 2016-10-27
Allow unique method on Collection and List

aheritier:


aheritier:
Please @reviewbybees

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -214,6 +214,8 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toList short[]
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods tokenize java.lang.String
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods tokenize java.lang.String java.lang.Character
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods tokenize java.lang.String java.lang.String
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods unique java.util.Collection
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods unique java.util.List
```
jglick: According to the test

    staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods unique java.util.List does not exist (or is an override)

aheritier: Should be fixed. Locally it is ok.

PR #94, opened 2016-10-27
More strenuous blacklisting

jglick:
UI continues that created in #24.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
@@ -167,7 +167,6 @@ class GroovyCallSiteSelector {
     }
 
     public static @CheckForNull Method staticMethod(@Nonnull Class<?> receiver, @Nonnull String method, @Nonnull Object[] args) {
-        // TODO should we check for inherited static calls?
         return findMatchingMethod(receiver, method, args);
     }
 
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
@@ -27,6 +27,7 @@ package org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.util.Arrays;
 import java.util.List;
 import javax.annotation.CheckForNull;
@@ -208,8 +209,7 @@ public abstract class EnumeratingWhitelist extends Whitelist {
                 }
             }
             try {
-                c.getDeclaredMethod(method, types(argumentTypes));
-                return true;
+                return !Modifier.isStatic(c.getDeclaredMethod(method, types(argumentTypes)).getModifiers());
             } catch (NoSuchMethodException x) {
                 return false;
             }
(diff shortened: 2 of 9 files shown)


jglick:
> Same thing for Hudson.getInstance

Yeah I just did not think people would be likely to use it—the blacklist is not intended to be an exhaustive list of all dangerous methods, merely a set of things that we suspect people might *try* to whitelist. But I can add it if you like.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -223,6 +223,13 @@ private boolean exists(Class<?> c, boolean start) throws Exception {
         @Override public String toString() {
             return "staticMethod " + signaturePart();
         }
+        @Override boolean exists() throws Exception {
+            try {
+                return Modifier.isStatic(type(receiverType).getDeclaredMethod(method, types(argumentTypes)).getModifiers());
```
jglick: Otherwise `StaticWhitelistTest.sanity` would fail, claiming `Hudson.getInstance` was an override. (`exists` is used only from tests so this is not a user-visible change.)

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelectorTest.java
```diff
@@ -74,6 +76,11 @@ public static void m2(int x) {}
         public static void m2(long x) {}
     }
 
+    @Test public void staticMethodsCannotBeOverridden() throws Exception {
+        assertEquals(Jenkins.class.getMethod("getInstance"), GroovyCallSiteSelector.staticMethod(Jenkins.class, "getInstance", new Object[0]));
+        assertEquals(Hudson.class.getMethod("getInstance"), GroovyCallSiteSelector.staticMethod(Hudson.class, "getInstance", new Object[0]));
```
jglick: Just making sure.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -670,6 +670,13 @@ public int getN() {
         public void explode() {}
     }
 
+    @Test public void staticMethodsCannotBeOverridden() throws Exception {
+        assertRejected(new StaticWhitelist(), "staticMethod jenkins.model.Jenkins getInstance", "jenkins.model.Jenkins.getInstance()");
+        assertRejected(new StaticWhitelist(), "staticMethod jenkins.model.Jenkins getInstance", "jenkins.model.Jenkins.instance");
+        assertRejected(new StaticWhitelist(), "staticMethod hudson.model.Hudson getInstance", "hudson.model.Hudson.getInstance()");
```
jglick: Ditto.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -167,7 +167,6 @@ private static boolean isInstancePrimitive(@Nonnull Class<?> type, @Nonnull Obje
     }
 
     public static @CheckForNull Method staticMethod(@Nonnull Class<?> receiver, @Nonnull String method, @Nonnull Object[] args) {
-        // TODO should we check for inherited static calls?
```
jglick: According to tests, no.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -208,8 +209,7 @@ private boolean exists(Class<?> c, boolean start) throws Exception {
                 }
             }
             try {
-                c.getDeclaredMethod(method, types(argumentTypes));
-                return true;
+                return !Modifier.isStatic(c.getDeclaredMethod(method, types(argumentTypes)).getModifiers());
```
jglick: Strengthening existing sanity check.

aheritier:
:bee:

PR #77, opened 2016-05-27
[FIXED JENKINS-35140] Added methods for JSON slurping in Groovy 2.x.

abayer:
https://issues.jenkins-ci.org/browse/JENKINS-35140

cc @reviewbybees @jglick

PR #98, opened 2016-11-24
Add signatures to whitelist

tnunes:
Would like to add these method signatures, which I believe are harmless, to the whitelist.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -16,6 +16,8 @@ staticMethod groovy.json.JsonOutput toJson java.util.Map
 staticMethod groovy.json.JsonOutput toJson java.util.UUID
 new groovy.json.JsonSlurper
 method groovy.json.JsonSlurper parseText java.lang.String
+new groovy.json.JsonSlurperClassic
+method groovy.json.JsonSlurperClassic parseText java.lang.String
 
 method groovy.lang.Binding hasVariable java.lang.String
 staticField groovy.lang.Closure DELEGATE_FIRST
@@ -107,6 +109,7 @@ method java.util.Collection addAll java.util.Collection
 method java.util.Collection contains java.lang.Object
 method java.util.Collection isEmpty
 method java.util.Collection size
+new java.util.Date
 new java.util.Date long
 method java.util.Iterator hasNext
 method java.util.Iterator next
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/jenkins-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/jenkins-whitelist
@@ -4,8 +4,6 @@ method hudson.model.AbstractBuild getEnvironments
 staticMethod hudson.model.Environment create hudson.EnvVars
 method hudson.model.Item getParent
 method hudson.model.ModelObject getDisplayName
-method hudson.model.Run getCause java.lang.Class
-method hudson.model.Run getCauses
 method hudson.model.Run getFullDisplayName
 method hudson.model.User getFullName
 method hudson.model.User getId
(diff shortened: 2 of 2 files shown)


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -16,6 +16,8 @@ staticMethod groovy.json.JsonOutput toJson java.util.Map
 staticMethod groovy.json.JsonOutput toJson java.util.UUID
 new groovy.json.JsonSlurper
 method groovy.json.JsonSlurper parseText java.lang.String
+new groovy.json.JsonSlurperClassic
```
jglick: Better not; see #77.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -107,6 +109,7 @@ method java.util.Collection addAll java.util.Collection
 method java.util.Collection contains java.lang.Object
 method java.util.Collection isEmpty
 method java.util.Collection size
+new java.util.Date
```
jglick: fine

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/jenkins-whitelist
```diff
@@ -4,6 +4,8 @@ method hudson.model.AbstractBuild getEnvironments
 staticMethod hudson.model.Environment create hudson.EnvVars
 method hudson.model.Item getParent
 method hudson.model.ModelObject getDisplayName
+method hudson.model.Run getCause java.lang.Class
+method hudson.model.Run getCauses
```
jglick: Hardly useful on its own, since you would need to then whitelist fields in the cause.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/jenkins-whitelist
```diff
@@ -26,3 +28,6 @@ field hudson.scm.EditType EDIT
 method hudson.scm.EditType getName
 method hudson.tools.ToolInstallation getHome
 method hudson.tools.ToolInstallation getName
+method org.jenkinsci.plugins.workflow.job.WorkflowRun getChangeSets
```
jglick: No, this is handled by `RunWrapper.getChangeSets()`.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/jenkins-whitelist
```diff
@@ -26,3 +28,6 @@ field hudson.scm.EditType EDIT
...
+method org.jenkinsci.plugins.workflow.support.steps.build.RunWrapper getRawBuild
```
jglick: To the contrary, this probably ought to be _blacklisted_.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/jenkins-whitelist
```diff
@@ -26,3 +28,6 @@ field hudson.scm.EditType EDIT
...
+method org.jenkinsci.plugins.workflow.support.steps.build.RunWrapper setDisplayName
```
jglick: This is already `@Whitelisted`.

tnunes:
Remove some signatures based on your feedback.

Regarding the `JsonSlurperClassic` and #77, I added both the constructor and the `parseText` method, similarly to how the `JsonSlurper` is whitelisted. Do you see any problem with that?

jglick:
I do not intend to support `JsonSlurper`. Use `jq` or similar command-line utilities as needed, or write a new Pipeline step that parses JSON to some `Serializable` types with no dynamic Groovy anything (preferably just `List`s and `Map`s and primitive wrappers).

hogarthj:
Can we perhaps split out the new date() approval?

I've already whitelisted this internally at my workplace but I imagine it'd be useful for others

jglick:
> Can we perhaps split out the new date() approval?

File a separate PR I suppose.

PR #99, opened 2016-12-01
Add all of the Collection type 'plus' methods

mkobit:


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist-groovy2
```diff
@@ -7,5 +7,28 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods each java.util.Map
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods each java.util.Set groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods each java.util.SortedSet groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods leftShift java.util.List java.lang.Object
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods plus java.lang.Iterable java.lang.Iterable
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods plus java.lang.Iterable java.lang.Object
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods plus java.lang.Object[] java.lang.Iterable
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods plus java.lang.Object[] java.lang.Object
```
jglick: This is in Groovy 1, so it should rather be in `generic-whitelist`.

mkobit: Should the Groovy 1 methods present in both 1 and 2 also be in the `generic-whitelist-groovy-2`?

jglick: No. Everything should be in `generic-whitelist` unless it is a method added to Groovy in Groovy 2.

mkobit: @jglick thanks, will take care of this tonight.

mkobit:
I think I've appropriately split up the methods between Groovy 1 and Groovy 2.

PR #105, opened 2017-02-06
[JENKINS-39159] If we create a URLClassLoader we should also close it

jglick:
[JENKINS-39159](https://issues.jenkins-ci.org/browse/JENKINS-39159)

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,8 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.6</version>
+    <version>2.21</version>
+    <relativePath/>
   </parent>
 
     <artifactId>script-security</artifactId>
@@ -13,8 +14,7 @@
     <description>Allows Jenkins administrators to control what in-process scripts can be run by less-privileged users.</description>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Script+Security+Plugin</url> 
     <properties>
-      <jenkins.version>1.580.1</jenkins.version>
-      <java.level>6</java.level>
+      <jenkins.version>1.625.3</jenkins.version>
     </properties>
     <licenses>
         <license>
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
@@ -145,6 +145,7 @@ public final class SecureGroovyScript extends AbstractDescribableImpl<SecureGroo
         if (!calledConfiguring) {
             throw new IllegalStateException("you need to call configuring or a related method before using GroovyScript");
         }
+        URLClassLoader urlcl = null;
         List<ClasspathEntry> cp = getClasspath();
         if (!cp.isEmpty()) {
             List<URL> urlList = new ArrayList<URL>(cp.size());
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -3,7 +3,8 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.6</version>
+    <version>2.21</version>
```
svanoort: Needed?

jglick: I update these as a matter of course.

jglick:
> what happens if we have a child classloader (returned) that depends on this now-closed parent?

Then any attempts to perform fresh class (or resource) loads from that child class loader will return errors.

Now consider what would need to happen. First, note that there is an implicit declared type for the return value: whatever the calling code casts it to, defined in the Java Platform or some Jenkins (core or plugin) code. Often this is an effectively final type like `Boolean` or `String`, or simply `Void` since the return value is ignored (the script may only call mutators). Suppose it is a type which the user script is intended to implement (not merely `abstract`—`Map` which would typically be a `HashMap`). This is largely hypothetical; the only example I can come up with is that an `Object` is returned and `toString` is called (implicitly).

Now suppose that the user specifies a classpath for their script, which is generally discouraged (this whole feature should be deprecated IMO). And the return value is of a type defined in that external JAR.

Merely returning a class defined in that JAR is fine—it will be loaded, `evaluate` will finish, and `toString` is called. So suppose in their `/home/jenkins/mystuff.jar` they had `mystuff/A.class`

```java
package mystuff;
public class A {
    @Override
    public String toString() {
        return B.print();
    }
}
```

and `mystuff/B.class`

```java
package mystuff;
class B {
    public static String print() {
        return "OK!";
    }
}
```

and the user script is

```groovy
new mystuff.A()
```

then the instance of `A` will be returned, and depending on the class loading behavior of the JVM, `B` might or might not yet have been resolved, so when the caller prints the return value you will get an error.

Compare this quite contrived problem to the very real `FileNotFoundException` reported by a user and apparently caused by failing to close a file handle.

svanoort:
@jglick Okay that makes sense, but let me narrow down the question: is there any circumstance besides a custom groovy classloader in which the GroovyShell would invoke the UrlClassLoader *after* the return here?  

If yes, then this is not safe and we need to create a cleanup thread that closes the classloader when we're done with the shell (or delegate to some cleanup/after-run hook in Groovy itself -- I do not know if there is such an animal or not). 

If no then this is quite a successful solution.

jglick:
> is there any circumstance besides a custom groovy classpath in which the GroovyShell would invoke the UrlClassLoader

We only create the `URLClassLoader` to begin with when you are using a custom Groovy classpath.

> in which the GroovyShell would invoke the UrlClassLoader _after_ the return here

In the (hopefully contrived) case above, the `URLClassLoader` would be used after the return from `evaluate`, as I mentioned. `GroovyShell` is not involved in that case. The class load is done by the JVM’s (somewhat) lazy class linking mechanism.

> when we're done with the shell

Unless the return value is of a type defined in the script itself, the `GroovyShell` should not be reachable once `evaluate` returns. Even if it is, and there is a custom classpath, it is unlikely the `URLClassLoader` would be needed after return. Since the Groovy compiler will resolve any type mentioned from the script, you would need another convoluted case to see such a problem:

```java
package mystuff;
public class A {
    public String whatever() {
        return B.something();
    }
}
```

```java
package mystuff;
public class B {
    public static String something() {
        return "late load";
    }
}
```

```groovy
class X {
    @Field a = new A()
    @Override def toString() {a.whatever()}
}
new X()
```

jglick:
> we're replacing a known issue with a _possible_ but unlikely issue

Right. This is the tricky bit with `URLClassLoader.close`: garbage collection is too haphazard to be relied on for an externally visible effect (file handle closure), yet lazy class linking makes it awkward to prove that all activity has ceased with some object. (For example this would be an obstacle to implementing dynamic plugin disable in Jenkins: some timer task involving novel plugin code could still be scheduled.)

PR #106, opened 2017-02-07
[FIXED JENKINS-41797] Make sure wrapper WeakHashMap will not be manipulated concurrenlty

olivergondza:
[JENKINS-41797](https://issues.jenkins-ci.org/browse/JENKINS-41797)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
@@ -28,6 +28,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.Whitelist;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


olivergondza:
I have started to see this repeatedly after moving an instance to Java 8. @jglick, as an author, can you have a look so we can get this released ASAP? Thanks.

olivergondza:
@reviewbybees

olivergondza:
> ...other than perhaps WeakHashMap occasionally going into an endless loop if mutated concurrently.

Yes that is the problem and it is a pretty severe one if it prevents Jenkins from loading the jobs altogether. This classifies as blocker IIRC.

Ad proper synchronization, I guess there are similar situations when different `WeakHashMap` operations can cause something similar. I am not familiar enough with the code to be confident about wrapping that into synchronized collection.

olivergondza:
> What makes you think synchronizing just this access is going to solve the problem?

What is happening IMO is that 2 loading threads get a race condition on one of the singleton whitelists and spin forever. The rest of the threads get there sooner or later and join the same party. If the access get synchronized, the race will never cause the livelock of the 2 threads.

olivergondza:
@jglick, so what do you suggest? Synchronize access to all the fields (as their consistency is not guaranteed with current implementation) which would resolve this bug as well?

jglick:
> so what do you suggest?

I think I suggested a fix above.

Certainly I could write that myself, but you seem to be the only one to have observed this issue (or at least the only one to report it), so I guess you could try a snapshot.

olivergondza:
[WS ignoring diff](https://github.com/jenkinsci/script-security-plugin/pull/106/files?w=1). Synchronizing on prominent final field instead of `this` not to expose the monitor.

olivergondza:
@jenkinsci/code-reviewers, please have a look.

olivergondza:
> This pull request originates from a CloudBees employee.

Huh? That must have been one heck of a weekend ...

jglick:
Hmm, the bots must be crazy.

PR #109, opened 2017-02-22
[JENKINS-42091] Add some needed methods of JSON superclass

nfalco79:
This PR is suggest by @rsandell [here](https://github.com/jenkinsci/pipeline-utility-steps-plugin/pull/22)


nfalco79:
Move to correct place so _sanity_ test pass

nfalco79:
The main reson of this PR is that when I get a JSON from JsonSlurper 
`def json = new JsonSlurper().from(file)`
or from step readJSON 
`def json = readJSON file: 'jsonFile'`
I do not know if the json instance is JSONObject or a JSONArray so i have to call isArray() method to understand how to access values or an exception is throwed.
```
if (json.isArray()) {
   echo json[0]
} else {
   echo json.key
}
```

I add to whitelist also JSON.size() because is a convenient method in case of JSONArray

nfalco79:
@reviewbybees

nfalco79:
maybe I used a wrong annotation :P

jglick:
Do not use `JsonSlurper` BTW.

PR #110, opened 2017-03-06
Revert minimal required version back to 1.609.3

olivergondza:
It is great #106 got released but I can not use it on 1.609. The codebase does not seem to rely on newer Jenkins version - Java 7 seems enough.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 target
 work
 
-.idea
+.*
+!.gitignore
+
 *.iml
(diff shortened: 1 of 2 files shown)


olivergondza:
Well, I do not feel strongly any longer as I have plugged the hole by inhouse patch. This plugin is mandatory dependnecy for matrix plugin and with 24 CPU cores halve of the boots got stuck for me.

jglick:
If it would be helpful, there is no problem with creating branches based on older core versions to which various important bug fixes are backported. I do not much care to work on backports for things that old, but do not mind if you do it yourself. At least then you would have an official build from the update center rather than an internal patch.

PR #114, opened 2017-04-10
[FIXED JENKINS-34741] Allow new Class(foo: bar) to work.

abayer:
[JENKINS-34741](https://issues.jenkins-ci.org/browse/JENKINS-34741)

Adds logic derived from Groovy's MetaClassImpl#invokeConstructor
method to return the no-parameters constructor when no matching
declared constructor is found, there's one argument, and that one
argument is a Map.

cc @reviewbybees


jglick:
cf. #67

PR #123, opened 2017-06-01
Add in string format methods from Java and some utility formatting methods from Groovy

mkobit:


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -88,6 +88,8 @@ method java.lang.String toLowerCase java.util.Locale
 method java.lang.String toUpperCase
 method java.lang.String toUpperCase java.util.Locale
 method java.lang.String trim
+staticMethod java.lang.String format java.util.Locale java.lang.String java.lang.Object[]
```
jglick: Bad sorting, should be moved up to allow test to pass.

mkobit: My fault! Thought I got this one right - will fix it

PR #124, opened 2017-06-01
Adding whitelist methods needed for DefaultGroovyMethods

abayer:
Might be better organized, might be things here that are problems. `invokeMethod` was needed for `sum`, so I've removed `sum` from the tests I'll be pushing up over in `workflow-cps` in an hour or so.

cc @reviewbybees

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -20,7 +20,6 @@ method groovy.json.JsonSlurper parseText java.lang.String
 method groovy.lang.Binding hasVariable java.lang.String
 staticField groovy.lang.Closure DELEGATE_FIRST
 staticField groovy.lang.Closure DELEGATE_ONLY
-staticField groovy.lang.Closure DONE
 staticField groovy.lang.Closure OWNER_FIRST
 staticField groovy.lang.Closure OWNER_ONLY
 staticField groovy.lang.Closure TO_SELF
@@ -29,19 +28,13 @@ method groovy.lang.Closure call java.lang.Object[]
 method groovy.lang.Closure curry java.lang.Object
 method groovy.lang.Closure curry java.lang.Object[]
 method groovy.lang.Closure getDelegate
-method groovy.lang.Closure getDirective
-method groovy.lang.Closure getMaximumNumberOfParameters
 method groovy.lang.Closure getResolveStrategy
 method groovy.lang.Closure ncurry int java.lang.Object
 method groovy.lang.Closure ncurry int java.lang.Object[]
 method groovy.lang.Closure setDelegate java.lang.Object
 method groovy.lang.Closure setResolveStrategy int
 method groovy.lang.GString plus java.lang.String
-staticMethod groovy.lang.ListWithDefault newInstance java.util.List boolean groovy.lang.Closure
-staticMethod groovy.lang.MapWithDefault newInstance java.util.Map groovy.lang.Closure
 method groovy.lang.Script getBinding
-staticMethod groovy.util.GroovyCollections combinations java.util.Collection
-new groovy.util.PermutationGenerator java.util.Collection
 
 # Needed for Groovy Templates to emit output:
 method java.io.PrintWriter print java.lang.Object
@@ -57,9 +50,6 @@ staticMethod java.lang.Boolean valueOf java.lang.String
 method java.lang.CharSequence charAt int
 method java.lang.CharSequence length
 method java.lang.Class getName
-method java.lang.Class getSimpleName
-method java.lang.Class isArray
-method java.lang.Class isInstance java.lang.Object
 method java.lang.Comparable compareTo java.lang.Object
 method java.lang.Enum name
 method java.lang.Enum ordinal
@@ -99,7 +89,6 @@ method java.lang.String toUpperCase
 method java.lang.String toUpperCase java.util.Locale
 method java.lang.String trim
 staticMethod java.lang.String valueOf java.lang.Object
-staticMethod java.lang.System arraycopy java.lang.Object int java.lang.Object int int
 staticMethod java.lang.System currentTimeMillis
 method java.lang.Throwable getMessage
 new java.net.MalformedURLException java.lang.String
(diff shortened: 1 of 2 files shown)


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1,6 +1,169 @@
...
+method org.codehaus.groovy.runtime.callsite.BooleanClosureWrapper callForMap java.util.Map$Entry
+new groovy.util.PermutationGenerator java.util.Collection
+new java.util.ArrayList
+new java.util.ArrayList int
+new java.util.LinkedHashMap
+new org.codehaus.groovy.runtime.ReverseListIterator java.util.List
+new org.codehaus.groovy.runtime.callsite.BooleanClosureWrapper groovy.lang.Closure
+new org.codehaus.groovy.util.ArrayIterator java.lang.Object[]
+staticField groovy.lang.Closure DONE
+staticMethod com.cloudbees.groovy.cps.CpsDefaultGroovyMethods $addEntry__java_util_Map__java_lang_Object java.util.Map java.lang.Object
```
jglick: You should not need any of these.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1,6 +1,169 @@
...
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods tail java.util.List
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toList java.lang.Object[]
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toList java.lang.String
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods withDefault java.util.List groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods withDefault java.util.Map groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods withLazyDefault java.util.List groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimilarCollection java.util.Collection
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimilarCollection java.util.Collection int
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimilarMap java.util.Map
+staticMethod org.codehaus.groovy.runtime.InvokerHelper getMetaClass java.lang.Object
```
jglick: Sounds dangerous, perhaps. What needs this?

jglick:
FTR this is a follow-up to https://github.com/jenkinsci/workflow-cps-plugin/pull/124.

abayer:
Will fix `sanity`. And am trying with the others removed now.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
abayer: Nothing, it looks like, but still checking.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
abayer: Turns out we do! Don't know why.

abayer: Looks like these are all private methods, fwiw.

abayer:
@jglick Ok, so the private methods are still needed but otherwise I think we're clear here.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1,4 +1,13 @@
-# groovy-cps
+# groovy-cps - not needed from workflow-cps 2.33+
+staticMethod com.cloudbees.groovy.cps.CpsDefaultGroovyMethods $addEntry__java_util_Map__java_lang_Object java.util.Map java.lang.Object
```
jglick: Why are we _adding_ to this list, if 2.32- had only `each`, and 2.33+ looks at the natural method signature? I am not following.

abayer: I'll try removing that one.

abayer:
@jglick What's special about the private methods is that they have to be white listed in their `CpsDefaultGroovyMethods` forms. Otherwise it doesn't work.

jglick:
Right, got that, but the question is why. Perhaps the `CpsWhitelist` code needs to scan `getDeclaredMethods` rather than just `getMethods`?

abayer:
@jglick That sounds right, yeah. I'll try it.

abayer:
Hrm, nope, `getDeclaredMethods()` makes everything worse. Experimenting further. =)

abayer:
@jglick Ok, https://github.com/jenkinsci/workflow-cps-plugin/pull/139 now updated to automatically pass for private methods in the category. That seems the best approach I can come up with.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -213,16 +262,34 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods getAt java.util.Li
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods getAt java.util.Map java.lang.Object
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods getAt java.util.regex.Matcher int
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods getChars java.lang.String
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods groupBy java.util.Collection groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods groupBy java.util.Collection java.lang.Object[]
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods groupBy java.util.Map groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods groupBy java.util.Map java.lang.Object[]
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods groupEntriesBy java.util.Map groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods inject java.lang.Object groovy.lang.Closure
```
jglick: Reminder to self: this is not at all as scary as it sounds! More or less `foldl` or similar.

Speaking of which, fun party trick: http://foldl.com/ (alas, seems to be dead now)

abayer:
@jglick We need a release of `script-security` now, I think.

PR #128, opened 2017-06-24
Jenkinsfile and POM updates

jglick:
Pending [INFRA-1247](https://issues.jenkins-ci.org/browse/INFRA-1247).

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.21</version>
+    <version>2.30</version>
     <relativePath />
   </parent>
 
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.21</version>
+    <version>2.30</version>
```
abayer: Shouldn't that be `2.31`?

jglick: Well, predated that POM. I did not bother to update it further at this time; surely that is not required.

jglick:
For some reason the repository is now recognized, but the last commit was not. Maybe need to wait another day.

pom.xml
abayer: Fair enough.

jglick:
@reviewbybees done

(waiting for an actual CI build before merging)

PR #130, opened 2017-07-04
Update jenkins-whitelist to allow timeout cause detection

idij:
@jglick

Per this example: [Pipeline: How to add an input step, with timeout, that continues if timeout is reached, using a default value](https://support.cloudbees.com/hc/en-us/articles/226554067-Pipeline-How-to-add-an-input-step-with-timeout-that-continues-if-timeout-is-reached-using-a-default-value), the only way to distinguish a timeout from an input dialog abort, from a system abort is by examining the causes, but doing so causes whitelist violations.

Ideally both the timeout and the input dialog abort would raise different subclassed exceptions, but while they're raising the FlowInterruptedException there doesn't seem to be another way than examining the causes, and so, IMHO, the methods needed to do so should be whitelisted.


idij:
I agree that it's not the best way, but it's the way the timeouts and aborts are implemented today that causes it.

As mentioned, I think sub-classed exceptions should be used as there's not much point in having a single  exception be re-used if it can't be queried to differentiate the cause.

Anyway, I'll close this PR.

PR #131, opened 2017-07-11
Added some static methods returning File to the blacklist

jglick:
Noticed in [JENKINS-45308](https://issues.jenkins-ci.org/browse/JENKINS-45308) that somebody was trying something which should never be allowed in the sandbox.

@reviewbybees

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/blacklist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/blacklist
@@ -20,9 +20,12 @@ staticMethod hudson.model.User get java.lang.String boolean java.util.Map
 staticMethod hudson.model.User getAll
 
 # Raw file operations could be used to compromise the Jenkins master.
+staticMethod java.io.File createTempFile java.lang.String java.lang.String
+staticMethod java.io.File createTempFile java.lang.String java.lang.String java.io.File
 new java.io.File java.lang.String
 new java.io.File java.lang.String java.lang.String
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/blacklist
```diff
@@ -20,9 +20,12 @@ staticMethod hudson.model.User get java.lang.String boolean java.util.Map
 staticMethod hudson.model.User getAll
 
 # Raw file operations could be used to compromise the Jenkins master.
+staticMethod java.io.File createTempFile java.lang.String java.lang.String
+staticMethod java.io.File createTempFile java.lang.String java.lang.String java.io.File
 new java.io.File java.lang.String
 new java.io.File java.lang.String java.lang.String
 new java.io.File java.net.URI
+staticMethod java.io.File listRoots
```
oleg-nenashev: I see no problem with that method if the Admin really wants to allow it

jglick: Not per se, nor indeed with the `File` constructors. My tactic is to prevent `File`s from being created in the first place, so we do not need to exhaustively blacklist all the things you might do with them. And since there is no legitimate reason for working with `File`s in a sandbox, I stand by this.

oleg-nenashev: okay

PR #134, opened 2017-07-14
[JENKINS-37129] unclassified method vs MethodMissingException

kohsuke:
[JENKINS-37129](https://issues.jenkins-ci.org/browse/JENKINS-37129)

As an user, I don't always precisely remember what methods an object has, an I often end up making mistakes. From workflow-cps perspective, this results in trying to call a method that does not exist.

This must result in `MethodMissingException`, just like normal Groovy interpreter. The code currently makes this a `RejectedAccessException`, making the user believe that the method exists but the access it not allowed.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
@@ -26,6 +26,7 @@ package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
 
 import groovy.lang.GroovyRuntimeException;
 import groovy.lang.MetaMethod;
+import groovy.lang.MissingMethodException;
 import groovy.lang.MissingPropertyException;
 import groovy.lang.Script;
 import hudson.Functions;
@@ -110,7 +111,8 @@ final class SandboxInterceptor extends GroovyInterceptor {
                 return super.onMethodCall(invoker, receiver, method, args);
             }
 
-            throw new RejectedAccessException("unclassified method " + EnumeratingWhitelist.getName(receiver.getClass()) + " " + method + printArgumentTypes(args));
+            // no such method exists
+            throw new MissingMethodException(method, receiver.getClass(), args);
         } else if (whitelist.permitsMethod(m, receiver, args)) {
             return super.onMethodCall(invoker, receiver, method, args);
         } else if (method.equals("invokeMethod") && args.length == 2 && args[0] instanceof String && args[1] instanceof Object[]) {
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
@@ -24,10 +24,6 @@
 
 package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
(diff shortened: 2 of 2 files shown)


kohsuke:
@reviewbybees

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -698,4 +699,29 @@ private static void assertRejected(Whitelist whitelist, String expectedSignature
         }
     }
 
+    @Issue("JENKINS-37129")
+    @Test public void methodMissingException() throws Exception {
+        // the case where the unsafe receiver type causes the security check to fail
+        try {
+            assertEvaluate(new GenericWhitelist(), "should fail", "[].noSuchMethod()");
+        } catch (MissingMethodException e) {
```
jglick: Will need to be reworked slightly after #132.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -698,4 +699,29 @@ private static void assertRejected(Whitelist whitelist, String expectedSignature
...
+            assertEquals(e.getType(),ArrayList.class);
+            assertThat(e.getMethod(),is("noSuchMethod"));
+        }
+
+        // trying to call an existing method that's not safe
+        try {
+            assertEvaluate(new GenericWhitelist(), "should fail", "[].class.classLoader");
+        } catch (RejectedAccessException e) {
+            assertEquals("method java.lang.Class getClassLoader", e.getSignature());
```
jglick: This is already covered by `assertRejected`.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -698,4 +699,29 @@ private static void assertRejected(Whitelist whitelist, String expectedSignature
...
+        }
+
+        // the case where the safe receiver type causes the security check to pass
```
jglick: What is the difference between a “safe” and “unsafe” receiver type? I am not following the distinction between the first and third cases here.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -110,7 +111,8 @@
                 return super.onMethodCall(invoker, receiver, method, args);
             }
 
-            throw new RejectedAccessException("unclassified method " + EnumeratingWhitelist.getName(receiver.getClass()) + " " + method + printArgumentTypes(args));
+            // no such method exists
+            throw new MissingMethodException(method, receiver.getClass(), args);
```
jglick: What about `MissingFieldException` and `MissingPropertyException`? (I do not think `MissingClassException` is relevant here.)

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
kohsuke: Prior to the fix, I noticed that the error message was different between those two cases. `1.noSuchMethod()` already result in `MissingMethodException`, and I assumed it was because the receiver type `Integer` is already whitelisted so the sandbox invoker gets a green light to attempt to invoke a method.

`[].noSuchMethod()` failed differently, presumably because whitelist rule rejects this.

jglick: Misanalyzed; corrected in #135.

PR #137, opened 2017-08-01
Add missing whitelist entries for DefaultGroovyMethods sort and contains operations on arrays

svanoort:
Plus containsAll on collections.


svanoort:
@reviewbybees especially @jglick and @abayer

svanoort:
@jglick should resolve the failure (looks like I made a bad assumption about how the sort ordering handles capitalization).

jglick:
@reviewbybees done

PR #141, opened 2017-08-16
[JENKINS-46191] Verify fix from groovy-sandbox

abayer:
[JENKINS-46191](https://issues.jenkins-ci.org/browse/JENKINS-46191)

Mainly here to verify that nothing else is broken by the upstream change in https://github.com/jenkinsci/groovy-sandbox/pull/39, but no harm in duplicating the test from upstream as well.

cc @reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
@@ -859,4 +859,10 @@ public class SandboxInterceptorTest {
         assertRejected(new StaticWhitelist(), "staticMethod jenkins.model.Jenkins getInstance", "jenkins.model.Jenkins x = jenkins.model.Jenkins.getInstance()");
     }
 
+    @Issue("JENKINS-46191")
+    @Test
+    public void emptyDeclaration() throws Exception {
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -859,4 +859,10 @@ public void rhsOfDeclarationTransformed() throws Exception {
         assertRejected(new StaticWhitelist(), "staticMethod jenkins.model.Jenkins getInstance", "jenkins.model.Jenkins x = jenkins.model.Jenkins.getInstance()");
     }
 
+    @Issue("JENKINS-46191")
+    @Test
+    public void emptyDeclaration() throws Exception {
+        assertEvaluate(new GenericWhitelist(), "abc", "String a; a = 'abc'; return a");
+    }
```
recampbell: What about covering other pattern such as declaration and initialization together? Assignment of one variable to another?

Are there test cases we can steal from groovy itself?

abayer: Those I believe we already have. But yeah, I need to spend some time locking in more thorough coverage.

abayer: (yeah, we've got those two cases well covered in various places)

PR #143, opened 2017-08-21
Increasing test coverage of Groovy syntax

abayer:
With some of the regressions recently, we probably would have caught them if we'd had better coverage of the language as a whole. So, let's add more coverage. This has already uncovered one bug, [JENKINS-46327](https://issues.jenkins-ci.org/browse/JENKINS-46327), and at least one fairly simple operator call that needed a whitelisting (`3.0 * 5.0`). More to come...

cc @reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyLanguageCoverageTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyLanguageCoverageTest.java
@@ -1,0 +1,152 @@
+package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
+
+import org.jenkinsci.plugins.scriptsecurity.sandbox.Whitelist;
+import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.GenericWhitelist;
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ErrorCollector;
+import org.jvnet.hudson.test.Issue;
+
+public class GroovyLanguageCoverageTest {
+
+    @Rule
+    public ErrorCollector errors = new ErrorCollector();
+
+    private void assertEvaluate(Whitelist whitelist, Object expected, String script) {
+        SandboxInterceptorTest.assertEvaluate(whitelist, expected, script, errors);
+    }
+
+    @Ignore("This fails on m.\"fruit${bKey}\" returning null due to JENKINS-46327")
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyLanguageCoverageTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+import org.jenkinsci.plugins.scriptsecurity.sandbox.Whitelist;
+import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.GenericWhitelist;
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ErrorCollector;
+import org.jvnet.hudson.test.Issue;
+
+public class GroovyLanguageCoverageTest {
```
jglick: Some stuff could probably be moved here from `SandboxInterceptorTest` BTW.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyLanguageCoverageTest.java
```diff
@@ -0,0 +1,152 @@
...
+                        "assert lazyGString.toString() == 'value == 1'\n" +
+                        "number = 2\n" +
+                        "assert eagerGString == 'value == 1'\n" +
+                        "assert lazyGString == 'value == 2'\n" +
+                        "return true\n"
+        );
+    }
+
+    @Test
+    public void arithmeticOperators() throws Exception {
```
jglick: so this overlaps with some existing test case(s)

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyLanguageCoverageTest.java
abayer: Was debating that but decided at least for now to keep these tests all in one place as I add them.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyLanguageCoverageTest.java
abayer: I'm ok with that - I'd rather some redundancy than end up missing some coverage because I misremembered what was covered elsewhere.

bitwiseman:
@abayer Any reason not to merge this?

PR #145, opened 2017-08-22
[JENKINS-46358] Whitelist new StringGroovyMethods

abayer:
[JENKINS-46358](https://issues.jenkins-ci.org/browse/JENKINS-46358)

These will be added to CPS in https://github.com/cloudbees/groovy-cps/pull/70

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
@@ -45,6 +45,7 @@ import org.codehaus.groovy.runtime.DefaultGroovyMethods;
 import org.codehaus.groovy.runtime.EncodingGroovyMethods;
 import org.codehaus.groovy.runtime.ProcessGroovyMethods;
 import org.codehaus.groovy.runtime.SqlGroovyMethods;
+import org.codehaus.groovy.runtime.StringGroovyMethods;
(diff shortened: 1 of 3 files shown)


svanoort:
Looks like the sanity check is failing -- something about methods not existing?  I'm assuming that's an issue with the sanity check itself and the new rules are fine though

abayer:
@svanoort Nah, they were legit sanity failures, which I've now cleaned up.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -93,17 +95,25 @@
...
-                    } else {
-                        throw StaticWhitelist.rejectStaticMethod(dgmMethod);
+                    } else if (foundDgmMethod == null) {
+                        foundDgmMethod = dgmMethod;
                     }
                 }
             }
 
+            // Some methods are found by GroovyCallSiteSelector in both DefaultGroovyMethods and StringGroovyMethods, so
+            // we're iterating over the whole list before we decide to fail out on the first failure we found.
```
jglick: Makes sense, I think, but would be nice to have a test case proving that this change is effective.

abayer: Good point. I've got that tested over in `workflow-CPS` but not here.

PR #147, opened 2017-09-06
Reduce an unactionable log-level to DEBUG rather than WARNING

rtyler:
I have seen this a number of times in Code Valet and on ci.jenkins.io, and I
cannot find anything actionable about it, which makes me think it's likely more
useful in a debugging context (if at all).

It seems that this is only an indication that my Jenkins master(s) are under
severe CPU load when running a Pipeline

See also:
<https://github.com/jenkinsci/script-security-plugin/commit/7aa1994a0ea2087b44c5a84e0ef89749bedfdf02#commitcomment-24064166>

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
@@ -87,7 +87,7 @@ class SandboxResolvingClassLoader extends ClassLoader {
                             t.setName(origName);
                             long ms = (System.nanoTime() - start) / 1000000;
                             if (ms > 250) {
-                                LOGGER.log(Level.DEBUG, "took {0}ms to load/not load {1} from {2}", new Object[] {ms, name, parentLoader});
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
```diff
@@ -87,7 +87,7 @@
                             t.setName(origName);
                             long ms = (System.nanoTime() - start) / 1000000;
                             if (ms > 250) {
-                                LOGGER.log(Level.WARNING, "took {0}ms to load/not load {1} from {2}", new Object[] {ms, name, parentLoader});
+                                LOGGER.log(Level.DEBUG, "took {0}ms to load/not load {1} from {2}", new Object[] {ms, name, parentLoader});
```
jglick: No such level. I suppose you meant `FINE` or something.

rtyler:
@jglick For me as an administrator there is absolutely nothing this logging incites me to do, it's a junk log message without some means of rectiying the situation (summoning Sam is not an option).

I also have an allergy to completely arbitrary magic numbers like 250ms. 😄

rtyler:
This is no longer relevant, somebody else lowerd this to INFO

dwnusbaum:
It was lowered to INFO by PR #222 (just commenting for the backlink)

PR #148, opened 2017-09-07
Add Reader.splitEachLine

jakub-bochenski:
http://docs.groovy-lang.org/latest/html/groovy-jdk/java/io/Reader.html#splitEachLine(java.lang.String,%20groovy.lang.Closure)

Per @jglick
> For simple missing entries in the default whitelist I do not bother with JIRA—just file PRs.

Otherwise we are getting `org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods splitEachLine java.lang.String java.lang.String groovy.lang.Closure` with  script-security-plugin 1.33

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -196,7 +196,6 @@ method net.sf.json.JSON size
 
 # More Groovy:
 staticMethod org.codehaus.groovy.runtime.DateGroovyMethods format java.util.Date java.lang.String
-
 staticMethod org.codehaus.groovy.runtime.DateGroovyMethods format java.util.Date java.lang.String java.util.TimeZone
 staticMethod org.codehaus.groovy.runtime.DateGroovyMethods getDateTimeString java.util.Date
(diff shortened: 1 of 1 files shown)


jakub-bochenski:
`java.lang.AssertionError: Whitelist is out of order!  Please see issues above.`
What is the correct order of entries?

jakub-bochenski:
related: https://issues.jenkins-ci.org/browse/JENKINS-46747

abayer:
Correct order is alphabetical by package/class/method name. But yeah, this should probably wait for the [JENKINS-46747](https://issues.jenkins-ci.org/browse/JENKINS-46747) fix, which I'll probably handle next week (and which will include adding the appropriate whitelisting).

jglick:
Test failure, please fix.

jakub-bochenski:
@jglick done, but I think this actually needs to be fixed in code as per https://github.com/jenkinsci/script-security-plugin/pull/148#issuecomment-328134731

jglick:
> this should probably wait for the JENKINS-46747 fix

That only applies to Pipeline. The method can be used from other scripts.

svanoort:
Thanks for the contribution @jakub-bochenski - I'm merging now

PR #149, opened 2017-09-08
[JENKINS-46757] Reproduced in test

jglick:
[JENKINS-46757](https://issues.jenkins-ci.org/browse/JENKINS-46757)

@reviewbybees


jglick:
Analyzed in JIRA already. Probably not worth fixing.

@reviewbybees done

PR #155, opened 2017-09-28
[JENKINS-47159] Catch GroovyCastException to enable finding the correct constructor when

antweiss:
varargs is involved.

Otherwise we're hitting [JENKINS-47159](https://issues.jenkins-ci.org/browse/JENKINS-47159)

This allows skipping the exception in trying to cast list to varargs and allows to find the correct constructor.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
@@ -100,11 +100,17 @@ class GroovyCallSiteSelector {
                 // not a varargs call
                 return parameters;
             } else {
-                Object array = DefaultTypeTransformation.castToVargsArray(parameters, 0, parameterTypes[fixedLen]);
-                Object[] parameters2 = new Object[fixedLen + 1];
-                System.arraycopy(parameters, 0, parameters2, 0, fixedLen);
-                parameters2[fixedLen] = array;
-                return parameters2;
+                try { //if casting  is unsuccessful - just return as is
+                    Object array = DefaultTypeTransformation.castToVargsArray(parameters, 0, parameterTypes[fixedLen]);
+                    Object[] parameters2 = new Object[fixedLen + 1];
+                    System.arraycopy(parameters, 0, parameters2, 0, fixedLen);
+                    parameters2[fixedLen] = array;
+                    return parameters2;
+                }
+                catch(org.codehaus.groovy.runtime.typehandling.GroovyCastException e)
+                {
+                    return parameters;
+                }
             }
         } else {
             return parameters;
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelectorTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelectorTest.java
@@ -32,6 +32,8 @@ import hudson.EnvVars;
 import hudson.model.Hudson;
 import java.io.PrintWriter;
 import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Map;
 import jenkins.model.Jenkins;
 import org.codehaus.groovy.runtime.GStringImpl;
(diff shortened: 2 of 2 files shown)


antweiss:
In general I don't think this is the right way to fix this. 
But at the very least - it doesn't break current functionality.
What currently happens is if the cast succeeds (an example is when EnvVars(Map<string, String>) is called and tested against EnvVars(String ...vars)  - the map is converted to a string )  - matches() returns true. Even when shouldn't. But then the constructor iteration doesn't exit - it continues and finds the correct constructor.  This works when the varargs constructor is first in the list - which is what seems to be happening in practice but isn't promised in any way by getDeclaredConstructors()
But what if it's not? 
Seems like a wrong constructor would be called.

abayer:
Ow, that's annoying. Let me think about this a bit to see if I can come up with a more "right" way to solve this, and if not, we'll go ahead with this approach. Thanks!

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -100,11 +100,17 @@ private static boolean matches(@Nonnull Class<?>[] parameterTypes, @Nonnull Obje
                 // not a varargs call
                 return parameters;
             } else {
-                Object array = DefaultTypeTransformation.castToVargsArray(parameters, 0, parameterTypes[fixedLen]);
-                Object[] parameters2 = new Object[fixedLen + 1];
-                System.arraycopy(parameters, 0, parameters2, 0, fixedLen);
-                parameters2[fixedLen] = array;
-                return parameters2;
+                try { //if casting  is unsuccessful - just return as is
```
abayer: Ok...so what would be ideal here is if we could determine the base type shared by all the parameters, i.e., in the example in [JENKINS-47159](https://issues.jenkins-ci.org/browse/JENKINS-47159), that'd be `ParameterValue`. Let me see what I can do there.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelectorTest.java
```diff
@@ -92,6 +94,10 @@ public static void m2(long x) {}
     @Issue("JENKINS-45117")
     @Test public void constructorVarargs() throws Exception {
         assertEquals(EnvVars.class.getConstructor(), GroovyCallSiteSelector.constructor(EnvVars.class, new Object[0]));
+        Map<String, String> myMap = new HashMap<String, String>();
+        myMap.put("ONE", "one");
+        myMap.put("TWO", "two"); 
+        assertEquals(EnvVars.class.getConstructor(Map.class), GroovyCallSiteSelector.constructor(EnvVars.class, new Object[] {myMap}));
```
abayer: This passes for me without the change to `GroovyCallSiteSelector`, so I'm not sure whether that change is actually relevant?

abayer:
@antweiss I think I've got a better fix over at https://github.com/jenkinsci/script-security-plugin/pull/156

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelectorTest.java
antweiss: Are you referring to the change in test? I know - it succeeds exactly becase of what I explained in my comment.
Do you think I should replace it with a test that uses PrarametersAction.class (which is what the fix was intended for?) It makes sense. I just didn't want to add additional dependencies.

abayer: See my change over in #156 - I think it's better. =)

antweiss: Yes. looks like the proper way to fix this. 
Will build and test with our setup to verify.

antweiss:
@abayer thanks - verified it fixes our issue

abayer:
Closed in favor of #156

PR #161, opened 2017-11-01
[JENKINS-47758] Script security provides automatic memory leak protection to many groovy scripts

svanoort:
[JENKINS-47758](https://issues.jenkins-ci.org/browse/JENKINS-47758)

Snapshot deployed as 1.35-20171101.213128-3 for testing
# TODO:

- [x] Locally test workflow-cps with updated script-security - PASSED
- [x] Locally test workflow-basic-steps with updated script-security - PASSED
- [x] Test workflow-global-libs with updated script-security - requires setting workflow-global-libs to core 2.7.3
- [x] Manually test against a generic set of Pipelines to verify it doesn't introduce any downstream issues  (scalability lab)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.30</version>
+    <version>2.35</version>
     <relativePath />
   </parent>
 
@@ -55,5 +55,29 @@
         </exclusion>
       </exclusions>
     </dependency>
-    </dependencies>
+    <dependency>
+      <groupId>org.jvnet.hudson.plugins</groupId>
+      <artifactId>groovy-postbuild</artifactId>
+      <version>2.3.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency> <!-- For memory leak testing. -->
+      <groupId>org.jvnet.hudson.plugins</groupId>
+      <artifactId>groovy-postbuild</artifactId>
+      <version>2.3.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency> <!-- Required for groovy postbuild. -->
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>matrix-project</artifactId>
+      <version>1.4.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency> <!-- Required for groovy postbuild, because needed for matrix-project after unbundling. -->
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>junit</artifactId>
+      <version>1.15</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.35</version>
+    <version>2.37</version>
     <relativePath />
   </parent>
 
@@ -54,12 +54,6 @@
           <artifactId>groovy</artifactId>
         </exclusion>
       </exclusions>
-    </dependency>
-    <dependency>
-      <groupId>org.jvnet.hudson.plugins</groupId>
-      <artifactId>groovy-postbuild</artifactId>
-      <version>2.3.1</version>
-      <scope>test</scope>
     </dependency>
     <dependency> <!-- For memory leak testing. -->
       <groupId>org.jvnet.hudson.plugins</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.30</version>
+    <version>2.35</version>
```
jglick: 2.37 is latest IIRC

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -159,23 +307,81 @@ public Object evaluate(ClassLoader loader, Binding binding) throws Exception {
         }
         try {
             loader = GroovySandbox.createSecureClassLoader(loader);
+
+            Field loaderF = GroovyShell.class.getDeclaredField("loader");
```
jglick: You need to catch `NoSuchFieldException` and quietly fall back gracefully, in case this code is refactored in a future version of Groovy. (I do _not_ recommend swallowing other reflection exceptions—log them as `WARNING`s: they are not expected. I.e., do not repeat this mistake: https://github.com/jenkinsci/cloudbees-folder-plugin/pull/79#discussion_r144733904)

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyMemoryLeakTest.java
```diff
@@ -0,0 +1,68 @@
...
+    public void loaderReleased() throws Exception {
+        FreeStyleProject p = r.jenkins.createProject(FreeStyleProject.class, "p");
+        p.addPublisher(new GroovyPostbuildRecorder(
+                new SecureGroovyScript(GroovyMemoryLeakTest.class.getName()+".register(this)", false, null),
+                2, false
+        ));
+        r.buildAndAssertSuccess(p);
+
+        assertFalse(LOADERS.isEmpty());
+        { // TODO it seems that the call to CpsFlowExecutionMemoryTest.register(Object) on a Script1 parameter creates a MetaMethodIndex.Entry.cachedStaticMethod.
```
jglick: This comment is presumably out of date.

svanoort:
@jglick Yes, I'd like to deduplicate the logic -- what you are describing is what I'd hoped to do (modulo some tweaks for extra cleanup that pipeline does). 

It might require providing some additional hooks for pipeline though because it does additional cleanup.  A problem for a later PR though.

svanoort:
Okay, looks ready for review.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -305,18 +305,29 @@ public Object evaluate(ClassLoader loader, Binding binding) throws Exception {
...
         try {
             loader = GroovySandbox.createSecureClassLoader(loader);
 
-            Field loaderF = GroovyShell.class.getDeclaredField("loader");
-            loaderF.setAccessible(true);
+            Field loaderF = null;
+            try {
+                loaderF = GroovyShell.class.getDeclaredField("loader");
+                loaderF.setAccessible(true);
+                canDoCleanup = true;
```
jglick: Or simply check `loaderF != null` later; then no `canDoCleanup` variable is needed.

svanoort:
@jglick  Since we know this is having a rather large impact in the wild and fixes 90% of issues, wouldn't it be better to go ahead and release if it looks reasonable, rather that trying to cover every single case?  Then we can inform future iterations of this by what we learn from that (for example we may need to broaden it to better cover unanticipated cases with non-pipeline Groovy).   

Since we have kept the implementation private, we have freedom to change it as needed in order to later expose this in a broader form.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyMemoryLeakTest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.lang.ref.WeakReference;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Level;
+
+import static org.junit.Assert.assertFalse;
+
+/**
+ * @author Sam Van Oort
```
jglick: Well…parts. :-) Never believe an `@author` tag and delete them when you find them.

svanoort: oops, okay removing that (autogenerated by IDE, sorry, definitely not trying to claim credit for your hard work)

PR #164, opened 2017-11-08
Removing test dependency on GroovyPostbuildRecorder

jglick:
Simplification of #161. Yes I checked that the test still fails if

```java
canDoCleanup = true;
```

is commented out.

@reviewbybees


recampbell:
@oleg-nenashev Please review

oleg-nenashev:
Retriggering CI

PR #165, opened 2017-11-08
Introduced SystemCommandLanguage

jglick:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/languages/SystemCommandLanguage.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/languages/SystemCommandLanguage.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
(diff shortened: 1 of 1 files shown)


jglick:
> Don't understand what this is or what it is intended to accomplish

Entirely at the service of the downstream PR.

> the downstream appears to be a brand new plugin with little-to-no documentation

It is only new as a plugin. As a feature of core it has existed for years. See https://github.com/jenkinsci/jenkins/pull/3076.

jglick:
See https://github.com/jenkinsci/jenkins/pull/3076/commits/e2a0792811a4a3adf07370738b4500996b3139c8 in particular.

jglick:
> Maybe it makes sense to distinguish Shell and Batch languages

This is neither. The string is tokenized but then executed as an argv. If you wish to use, say, Bourne Shell features, you would need to use `sh -c "…"`.

jglick:
@reviewbybees done

PR #166, opened 2017-11-20
Hide link to /scriptApproval when there is nothing approved or pending

jglick:
Makes this plugin behave like more of a pure library when it is not in active use: there is no particular reason to even see this management link if all attempted uses of script security (if any) have been successful uses of default whitelist entries in the sandbox.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.37</version>
+    <version>3.0</version>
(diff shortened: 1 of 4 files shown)


jglick:
@reviewbybees done

pom.xml
```diff
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.37</version>
+    <version>3.0</version>
```
jglick: cf. #167

PR #169, opened 2017-12-05
performant multi-run support

akostadinov:
Hello, I want to be able to run a groovy script many times without needless overhead.
This is needed for me to implement custom log lines processing in Logstash plugin.
At the same time I'd like the convenience and safety of all the magic presently done
by the secure groovy script plugin. This modification is to allow that.

Sorry about white-space changes. It's because of the editor settings. If you don't
want them, I'll go ahead and revert them. Same about the EditorConfig file. Just
thought it is a useful addition.
If you like to keep the removed needless whilte spaces, you could add `?w=1` to url
of the file diff to have them filtered out while reviewing.


akostadinov:
Any thoughts on this one? Would be really useful to know whether such functionality can be eventually accepted and whether this is the right approach for such.

Thank you.

jglick:
> If you don't want them, I'll go ahead and revert them.

Yes please. It is not just about comfort when reviewing diffs; if we ever need to `git cherry-pick` changes, etc., it is important for changes to be textually minimal.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java
```diff
@@ -144,6 +141,20 @@ public static Object run(@Nonnull Script script, @Nonnull final Whitelist whitel
         }
     }
 
+    /**
+     * Prepares a sandbox for a script.
+     * You must have used {@link #createSecureCompilerConfiguration} to prepare the Groovy shell.
+     * @param script a script ready to {@link Script#run}, created for example by {@link GroovyShell#parse(String)}
+     * @param whitelist the whitelist to use, such as {@link Whitelist#all()}
+     * @return the sandbox for running this script
```
jglick: Would be helpful to `@link` to `register` and `unregister` so callers can see how to use it.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -277,6 +280,10 @@ private static void cleanUpObjectStreamClassCaches(@Nonnull Class<?> clazz) thro
         }
     }
 
+    public PreparedScript prepare(ClassLoader loader, Binding binding) throws IllegalAccessException, IOException {
```
jglick: Deserves Javadoc at least for the `loader` parameter, which in `evaluate` is nontrivial.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -287,63 +294,99 @@ private static void cleanUpObjectStreamClassCaches(@Nonnull Class<?> clazz) thro
...
-                    memoryProtectedLoader = new CleanGroovyClassLoader(loader);
-                    loaderF.set(sh, memoryProtectedLoader);
-                }
-                return sh.evaluate(ScriptApproval.get().using(script, GroovyLanguage.get()));
+                return preparedScript.run();
             }
+        }
 
-        } finally {
+        public void cleanUp() throws IOException {
```
jglick: Better implement `Closeable` and make this the `close` method, so `PreparedScript` can be used in a Java 7+ `try` block.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -287,63 +294,99 @@ private static void cleanUpObjectStreamClassCaches(@Nonnull Class<?> clazz) thro
...
 
-            Field loaderF = null;
-            try {
-                loaderF = GroovyShell.class.getDeclaredField("loader");
-                loaderF.setAccessible(true);
-                canDoCleanup = true;
-            } catch (NoSuchFieldException nsme) {
-                LOGGER.log(Level.FINE, "GroovyShell fields have changed, field loader no longer exists -- memory leak fixes won't work");
+    /**
+     * Allows access to execue a script multiple times without preparation and clean-up overhead.
```
jglick: typo

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -287,63 +294,99 @@ private static void cleanUpObjectStreamClassCaches(@Nonnull Class<?> clazz) thro
...
                 }
+            } catch (Exception e) {
+                cleanUp();
+                throw e;
+            }
+        }
 
+        public Object run() throws Exception {
+            if (sandbox) {
+                scriptSandbox.register();
```
jglick: Note to self: `register` and `unregister` calls should be cheap enough that they contribute no significant overhead.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
```diff
@@ -802,5 +802,40 @@ public void testSandboxClassResolution() throws Exception {
...
+        SecureGroovyScript.PreparedScript script = sgs.configuringWithKeyItem().prepare(Jenkins.getInstance().getPluginManager().uberClassLoader, b);
+        try {
+            Object res = script.run();
+            assertTrue((int) res == 5);
+            assertTrue((int) b.getVariable("a") == 6);
+            res = script.run();
+            assertTrue((int) res == 6);
+            assertTrue((int) b.getVariable("a") == 7);
+        } catch (Exception e) {
+            script.cleanUp();
```
jglick: This should be in a `finally` block, not a `catch` block (or just use `Closeable` trick mentioned above).

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
```diff
@@ -802,5 +802,40 @@ public void testSandboxClassResolution() throws Exception {
...
```
jglick: ditto

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
akostadinov: strange that I did this, I'll check your `Closeable` suggestion

akostadinov:
Hello, updated according to your reasonable comments. Please let me know if you see further possible improvements (or if I missed anything).

One question - would the automatic PR test job deploy a timestamped snapshot or does it have to be deployed in some other way?

jglick:
> would the automatic PR test job deploy a timestamped snapshot

Afraid not.

> does it have to be deployed in some other way?

IIRC you just need a jenkins.io account, then I use a helper script:

```bash
mvn clean install source:jar deploy:deploy -DaltDeploymentRepository=maven.jenkins-ci.org::default::https://repo.jenkins-ci.org/snapshots/ -DskipTests
```

Plain `mvn deploy` will work well enough in most cases (the above handles some corner cases in Jenkins components released to OSSRH etc.).

jglick:
> would the automatic PR test job deploy a timestamped snapshot

BTW that has been superseded by JEP-305. [Docs](https://github.com/jenkinsci/incrementals-tools/blob/master/README.md#incrementals)

PR #170, opened 2017-12-06
[FIXED JENKINS-46764] Log a useful message when scriptApproval is malformed

abayer:
[JENKINS-46764](https://issues.jenkins-ci.org/browse/JENKINS-46764)

Still fails to actually load an `ApprovedWhitelist` but at least it now logs an error message.

cc @reviewbybees


src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -668,8 +668,14 @@ public synchronized RejectedAccessException accessRejected(@Nonnull RejectedAcce
     @Restricted(NoExternalUse.class) // implementation
     @Extension public static final class ApprovedWhitelist extends ProxyWhitelist {
         public ApprovedWhitelist() throws IOException {
-            reconfigure();
+            try {
+                reconfigure();
+            } catch (IOException e) {
+                LOG.log(Level.SEVERE, "Malformed signature entry in scriptApproval.xml: '" + e.getMessage() + "'");
+                throw e;
```
jglick: Why even rethrow?

It is not clear from either JIRA or the PR who was catching the `IOException` to begin with and where it was swallowed, but anyway I see no reason to rethrow this. Just delete and remove the `throws` clause.

PR #171, opened 2017-12-06
[FIXED JENKINS-46213] Treat trailing array params as varargs

abayer:
[JENKINS-46213](https://issues.jenkins-ci.org/browse/JENKINS-46213)

Groovy's method lookup logic will treat `sprintf(fmt, "string", "other
string")` the same as `sprintf(fmt, ["string", "other string"])`. We
should emulate that.

cc @reviewbybees


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -195,16 +195,50 @@ private static boolean isInstancePrimitive(@Nonnull Class<?> type, @Nonnull Obje
...
                 }
             }
         }
         return candidate;
     }
 
+    /**
+     * Emulates, with some tweaks, {@link org.codehaus.groovy.reflection.ParameterTypes#isVargsMethod(Object[])}
+     */
+    private static boolean isVargsMethod(Method m, Object[] args) {
```
jglick: for legibility prefer `varArgs` to `vargs`

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -195,16 +195,50 @@ private static boolean isInstancePrimitive(@Nonnull Class<?> type, @Nonnull Obje
 
     private static Method findMatchingMethod(Class<?> receiver, String method, Object[] args) {
         Method candidate = null;
+
         for (Method m : receiver.getDeclaredMethods()) {
-            if (m.getName().equals(method) && (matches(m.getParameterTypes(), args, m.isVarArgs()))) {
-                if (candidate == null || isMoreSpecific(m, m.getParameterTypes(), m.isVarArgs(), candidate, candidate.getParameterTypes(), candidate.isVarArgs())) {
+            if (m.getName().equals(method) && (matches(m.getParameterTypes(), args, isVargsMethod(m, args)))) {
+                if (candidate == null || isMoreSpecific(m, m.getParameterTypes(), isVargsMethod(m, args), candidate,
```
jglick: introduce variable for varargs check, now that it is no longer trivial

PR #172, opened 2017-12-08
Blacklist System.setProperty

jglick:
Noticed this whitelisted in someone’s `scriptApproval.xml`. No idea why, but cannot be a good idea.

@reviewbybees

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/blacklist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/blacklist
@@ -68,6 +68,9 @@ staticMethod java.lang.System getProperty java.lang.String java.lang.String
 staticMethod java.lang.System getenv
 staticMethod java.lang.System getenv java.lang.String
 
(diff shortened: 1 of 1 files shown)


jglick:
@reviewbybees done

svanoort:
Yeah, why not.  With caveat that we might have to revert if someone has a legit reason to approve it.

oleg-nenashev:
> Yeah, why not. With caveat that we might have to revert if someone has a legit reason to approve it.

Our blacklist is not a blacklist. They just warn the admin in the UI, but he can allow it if he feels brave

daniel-beck:
@svanoort Isn't the blacklist just adding a "Approving this will introduce a vulnerability, so you really shouldn't"?

daniel-beck:
I agree with @oleg-nenashev — users still can approve this but it's obviously not a good idea to do this. I can think of several system properties that will just wreck Jenkins.

svanoort:
Ah, I've never been silly enough to approve something in the blacklist, so I didn't know you could.  

Okay, merging then.

PR #173, opened 2017-12-12
[FIXED JENKINS-48501] Null-safety in varargs->array check

abayer:
[JENKINS-48501](https://issues.jenkins-ci.org/browse/JENKINS-48501)

Added some more null checks in the vicinity just to be safe, but the
main thing here is that we needed to avoid an NPE *and* treat a null
arg as if it were of the target method's trailing array parameter's
component type, since, well, null is every type.

Downstream test of the specific issue that uncovered this (variable declaration with a specified type in a CPS script without a right-hand side) coming momentarily.

cc @reviewbybees @rsandell

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
@@ -193,15 +193,17 @@ class GroovyCallSiteSelector {
         return findMatchingMethod(receiver, method, args);
     }
 
-    private static Method findMatchingMethod(Class<?> receiver, String method, Object[] args) {
+    private static Method findMatchingMethod(@Nonnull Class<?> receiver, @Nonnull String method, @Nonnull Object[] args) {
         Method candidate = null;
 
         for (Method m : receiver.getDeclaredMethods()) {
-            boolean isVarArgs = isVarArgsMethod(m, args);
-            if (m.getName().equals(method) && (matches(m.getParameterTypes(), args, isVarArgs))) {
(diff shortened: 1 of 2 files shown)


abayer:
Ah, repo.azure.jenkins.io outages are fun.

abayer:
Actually, looks like the real problem is nasty slowness on repo.jenkins-ci.org itself.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelector.java
```diff
@@ -193,15 +193,17 @@ private static boolean isInstancePrimitive(@Nonnull Class<?> type, @Nonnull Obje
...
+    private static Method findMatchingMethod(@Nonnull Class<?> receiver, @Nonnull String method, @Nonnull Object[] args) {
         Method candidate = null;
 
         for (Method m : receiver.getDeclaredMethods()) {
-            boolean isVarArgs = isVarArgsMethod(m, args);
-            if (m.getName().equals(method) && (matches(m.getParameterTypes(), args, isVarArgs))) {
-                if (candidate == null || isMoreSpecific(m, m.getParameterTypes(), isVarArgs, candidate,
-                        candidate.getParameterTypes(), isVarArgsMethod(candidate, args))) {
-                    candidate = m;
+            if (m != null) {
```
jglick: Huh? `getDeclaredMethods` cannot return an array containing nulls. I think you can revert this hunk.

abayer: I got paranoid. Will clean up.

PR #176, opened 2018-01-09
Improve UX in case of new dangerous signatures

Wadeck:
- add a button to only clear the dangerous signature that were approved instead of clearing everything
- add a test on the clear methods

~No associated ticket.~ Part of [JENKINS-22660](https://issues.jenkins-ci.org/browse/JENKINS-22660)

@reviewbybees @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
@@ -824,23 +824,19 @@ import org.kohsuke.stapler.bind.JavaScriptMethod;
     @JavaScriptMethod public synchronized String[][] clearDangerousApprovedSignatures() throws IOException {
         Jenkins.getInstance().checkPermission(Jenkins.RUN_SCRIPTS);
 
-        TreeSet<String> newApprovedSignatures = new TreeSet<>();
-        for (String signature : approvedSignatures) {
-            if(!StaticWhitelist.isBlacklisted(signature)){
-                newApprovedSignatures.add(signature);
+        Iterator<String> it = approvedSignatures.iterator();
+        while (it.hasNext()) {
+            if (StaticWhitelist.isBlacklisted(it.next())) {
+                it.remove();
             }
         }
-        approvedSignatures.clear();
-        approvedSignatures.addAll(newApprovedSignatures);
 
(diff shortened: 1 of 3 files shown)


jglick:
> No associated ticket.

Well, this is related to [JENKINS-22660](https://issues.jenkins-ci.org/browse/JENKINS-22660), though a full fix of that would require some more work.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -819,6 +819,32 @@ private synchronized void removePendingScript(String hash) {
...
+        Jenkins.getInstance().checkPermission(Jenkins.RUN_SCRIPTS);
+
+        TreeSet<String> newApprovedSignatures = new TreeSet<>();
+        for (String signature : approvedSignatures) {
+            if(!StaticWhitelist.isBlacklisted(signature)){
+                newApprovedSignatures.add(signature);
+            }
+        }
+        approvedSignatures.clear();
+        approvedSignatures.addAll(newApprovedSignatures);
```
jglick: Or more simply:

```java
Iterator<String> it = approvedSignatures.iterator();
while (it.hasNext()) {
    if (StaticWhitelist.isBlacklisted(it.next())) {
        it.remove();
    }
}
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/index.jelly
```diff
@@ -243,6 +248,11 @@ THE SOFTWARE.
                 You can also remove all previous signature approvals:
                 <button onclick="if (confirm('Really delete all approvals? Any existing scripts will need to be rerun and signatures reapproved.')) {clearApprovedSignatures()}">Clear Approvals</button>
             </p>
+            <j:if test="${!empty(dangerousApprovedSignatures)}">
+                Or you can just remove the dangerous ones:
+                <button onclick="if (confirm('Really delete all dangerous approvals? Any scripts that was using such signatures will need to be rerun and signatures reapproved.')) {clearDangerousApprovedSignatures()}">Clear only dangerous Approvals</button>
```
jglick: Certainly the confirmation message should be reworded, and possibly no confirmation is even needed for this case.

Wadeck:
Simplification done. Thank you for the review 👍

PR #179, opened 2018-01-25
Optimize whitelist check by using caching in ProxyWhitelist

svanoort:
We generate a canonical string handle for lookup describing the Constructor/method call/etc and use that so we don't have to do an exhaustive whitelist search. 

Not sure what if anything more is needed for tests, since the methods are fairly directly invoked -- open to ideas here?

After the patches in https://github.com/jenkinsci/structs-plugin/pull/30 and https://github.com/jenkinsci/structs-plugin/pull/31 this has the most lock contention in pipeline -- with 40 pipelines running at once, 'delegates' spends a lot of time locked and blocking other pipelines.

If we want people to be able to run lots of pipelines at once, this is a must-have.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
@@ -36,6 +36,8 @@ import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.GroovySandbox;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.ProxyWhitelist;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 
 /**
  * Determines which methods and similar members which scripts may call.
@@ -87,6 +89,7 @@ public abstract class Whitelist implements ExtensionPoint {
         return o == null ? "null" : getName(o.getClass());
     }
 
+    @Restricted(NoExternalUse.class)
     public static String[] argumentTypes(Class<?>[] argumentTypes) {
         String[] s = new String[argumentTypes.length];
         for (int i = 0; i < argumentTypes.length; i++) {
@@ -96,36 +99,44 @@ public abstract class Whitelist implements ExtensionPoint {
     }
 
     /** Canonical name for a field access. */
+    @Restricted(NoExternalUse.class)
     public static String canonicalFieldString(@Nonnull Field field) {
         return getName(field.getDeclaringClass()) + ' ' + field.getName();
     }
 
     /** Canonical name for a method call. */
+    @Restricted(NoExternalUse.class)
     public static String canonicalMethodString(@Nonnull Method method) {
         return joinWithSpaces(new StringBuilder(getName(method.getDeclaringClass())).append(' ').append(method.getName()), argumentTypes(method.getParameterTypes())).toString();
     }
 
     /** Canonical name for a constructor call. */
+    @Restricted(NoExternalUse.class)
     public static String canonicalConstructorString(@Nonnull Constructor cons) {
         return joinWithSpaces(new StringBuilder(getName(cons.getDeclaringClass())).append(' ').append(cons.getName()), argumentTypes(cons.getParameterTypes())).toString();
     }
 
+    @Restricted(NoExternalUse.class)
     public static String canonicalMethodSig(@Nonnull Method method) {
         return "method "+canonicalMethodString(method);
     }
 
+    @Restricted(NoExternalUse.class)
     public static String canonicalStaticMethodSig(@Nonnull Method method) {
         return "staticMethod "+canonicalMethodString(method);
     }
 
+    @Restricted(NoExternalUse.class)
     public static String canonicalConstructorSig(@Nonnull Constructor cons) {
         return "new "+canonicalConstructorString(cons);
     }
 
+    @Restricted(NoExternalUse.class)
     public static String canonicalFieldSig(@Nonnull Field field) {
         return "field "+canonicalFieldString(field);
     }
 
+    @Restricted(NoExternalUse.class)
     public static String canonicalStaticFieldSig(@Nonnull Field field) {
         return "staticField "+canonicalFieldString(field);
     }
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
@@ -100,9 +100,34 @@ public abstract class Whitelist implements ExtensionPoint {
         return getName(field.getDeclaringClass()) + ' ' + field.getName();
     }
 
-    /** Canonical name for a field access. */
+    /** Canonical name for a method call. */
     public static String canonicalMethodString(@Nonnull Method method) {
         return joinWithSpaces(new StringBuilder(getName(method.getDeclaringClass())).append(' ').append(method.getName()), argumentTypes(method.getParameterTypes())).toString();
+    }
+
+    /** Canonical name for a constructor call. */
+    public static String canonicalConstructorString(@Nonnull Constructor cons) {
+        return joinWithSpaces(new StringBuilder(getName(cons.getDeclaringClass())).append(' ').append(cons.getName()), argumentTypes(cons.getParameterTypes())).toString();
+    }
+
+    public static String canonicalMethodSig(@Nonnull Method method) {
+        return "method "+canonicalMethodString(method);
+    }
+
+    public static String canonicalStaticMethodSig(@Nonnull Method method) {
+        return "staticMethod "+canonicalMethodString(method);
+    }
+
+    public static String canonicalConstructorSig(@Nonnull Constructor cons) {
+        return "new "+canonicalConstructorString(cons);
+    }
+
+    public static String canonicalFieldSig(@Nonnull Field field) {
+        return "field "+canonicalFieldString(field);
+    }
+
+    public static String canonicalStaticFieldSig(@Nonnull Field field) {
+        return "staticField "+canonicalFieldString(field);
     }
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
@@ -66,6 +66,45 @@ public abstract class Whitelist implements ExtensionPoint {
 
     public abstract boolean permitsStaticFieldSet(@Nonnull Field field, @CheckForNull Object value);
 
+    // Utility methods for creating canonical string representations of the signature
+    public static final StringBuilder joinWithSpaces(StringBuilder b, String[] types) {
+        for (String type : types) {
+            b.append(' ').append(type);
+        }
+        return b;
+    }
+
+    public static @Nonnull String getName(@Nonnull Class<?> c) {
+        Class<?> e = c.getComponentType();
+        if (e == null) {
+            return c.getName();
+        } else {
+            return getName(e) + "[]";
+        }
+    }
+
+    public static @Nonnull String getName(@CheckForNull Object o) {
+        return o == null ? "null" : getName(o.getClass());
+    }
+
+    public static String[] argumentTypes(Class<?>[] argumentTypes) {
+        String[] s = new String[argumentTypes.length];
+        for (int i = 0; i < argumentTypes.length; i++) {
+            s[i] = getName(argumentTypes[i]);
+        }
+        return s;
+    }
+
+    /** Canonical name for a field access. */
+    public static String canonicalFieldString(@Nonnull Field field) {
+        return getName(field.getDeclaringClass()) + ' ' + field.getName();
+    }
+
+    /** Canonical name for a field access. */
+    public static String canonicalMethodString(@Nonnull Method method) {
+        return joinWithSpaces(new StringBuilder(getName(method.getDeclaringClass())).append(' ').append(method.getName()), argumentTypes(method.getParameterTypes())).toString();
+    }
+
     /**
      * Checks for all whitelists registered as {@link Extension}s and aggregates them.
      * @return an aggregated default list
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/CacheableWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/CacheableWhitelist.java
@@ -2,7 +2,8 @@ package org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists;
 
 /**
  * Marker interface indicating results of permission checks do not change unless the list of entries is modified
- *  and depend ONLY on the Field/Method/Constructor actually being invoked.
+ *  and depend ONLY on the Field/Method/Constructor actually being invoked, irrespective of classloader.
+ *  That means you must be able to represent the WhiteList entry with a String signature of className + method/field signature.
  * @author Sam Van Oort
  */
 public interface CacheableWhitelist {
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/CacheableWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/CacheableWhitelist.java
@@ -1,0 +1,9 @@
+package org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists;
+
+/**
+ * Marker interface indicating results of permission checks do not change unless the list of entries is modified
+ *  and depend ONLY on the Field/Method/Constructor actually being invoked.
+ * @author Sam Van Oort
+ */
+public interface CacheableWhitelist {
+}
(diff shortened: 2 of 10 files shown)


svanoort:
@reviewbybees @rsandell @abayer

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
```diff
@@ -66,6 +66,70 @@
 
     public abstract boolean permitsStaticFieldSet(@Nonnull Field field, @CheckForNull Object value);
 
+    // Utility methods for creating canonical string representations of the signature
+    public static final StringBuilder joinWithSpaces(StringBuilder b, String[] types) {
```
svanoort: `joinWithSpaces` and `getName` (both variants) and `argumentTypes` are all just relocated from EnumeratingWhitelist unchanged (allowing them to be reused here).

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -122,77 +155,168 @@ public ProxyWhitelist(Whitelist... delegates) {
 
     @Override public final boolean permitsMethod(Method method, Object receiver, Object[] args) {
         synchronized (this.delegates) {
-            for (Whitelist delegate : delegates) {
+            String sigString = Whitelist.canonicalMethodSig(method);
+            Boolean b = permittedCache.get(sigString);
```
rsandell: Is this really enough of a check? Only the method signature?. The actual ask is: is it allowed to run this method on __this__ object with __these__ arguments.

See for example [this implementation](https://github.com/jenkinsci/email-ext-plugin/blob/master/src/main/java/hudson/plugins/emailext/groovy/sandbox/MimeMessageInstanceWhitelist.java) for a bit unorthodox implementation.

svanoort: @rsandell So... here we're relying on a couple properties of the Whitelist.  First, for Whitelists implementing the `CacheableWhitelist` interface, they MUST only depend on the method/field being used and NOT on the object or arguments used.  Second, any Whitelist that permits the method means it can be used.  The combination means that we can cache an approval *if* it comes from a CacheableWhitelist. 

If the call/access is *not* permitted or was permitted by a non-Cacheable Whitelist, we know it doesn't come from a Cacheable whitelist and have to dynamically check it with each invocation, because the approval depends on the receiver or arguments.  But wait!  We can skip the CacheableWhitelist checks because know at that point that none of them granted permission, so *none of them are going to.* 

Make a bit more sense?

jglick:
Definitely deserves careful review—any mistakes here could be serious security issues.

svanoort:
@rsandell any chance of another peek, please?

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
jglick: Can this be package-private or `@Restricted`? I do not see why users of this API plugin would call it. Ditto most of the other newly added members.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -27,22 +27,58 @@
...
 import java.util.Map;
 import java.util.WeakHashMap;
 
 /**
  * Aggregates several whitelists.
+ *
+ * <p>To improve performance we split the {@link Whitelist}s into two categories:
+ *
+ * <ol>
+ *     <li>{@link CacheableWhitelist} for static whitelists whose permit/fail results can be cached.  This is because the lists do not change and depend ONLY on the signature of the Field/Method/Constructor (see the interface Javadocs).</li>
```
jglick: missing `@link`s I think

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -27,22 +27,58 @@
...
+ *     <li>Non-cacheable or "dynamic" Whitelists, which must be dynamically evaluated for each call, i.e. {@link AclAwareWhitelist}.</li>
+ * </ol>
+ *
+ * <p> We rely on the fact that an operation is permitted if ANY Whitelist approves it, and the majority of calls are approved by one of the CacheableWhitelist checks.
+ * By sorting the aggregated (delegate) Whitelists to evaluate CacheableWhitelists first, this allows us to partially cache lookups and use an optimized permission check.
+ *
+ * <ol>
+ *     <li>First, we generate a String signature specific to the Field/Method/Constructor call type</li>
+ *     <li>Then, this signature is used to check the cache {@link #permittedCache} - a stored "true" indicates that one of the CacheableWhitelist entries approved the method so it is always safe.</li>
```
jglick: Private member, so you cannot `@link` to it from public Javadoc.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -51,16 +87,37 @@
...
 
+    /** Sorter that ensures we check {@link CacheableWhitelist}s before the dynamically-evaluating whitelists such as
+     *  {@link AclAwareWhitelist}. */
+    static final Comparator<Whitelist> CACHEABLE_WHITELIST_FIRST = new Comparator<Whitelist>() {
+        @Override
+        public int compare(Whitelist o1, Whitelist o2) {
+            if (o1 instanceof CacheableWhitelist ^ o2 instanceof CacheableWhitelist) {  // If they don't match
+                return (o1 instanceof CacheableWhitelist) ? -1 : 1;
+            }
+            return 0;
```
jglick: Maybe more readable as

```java
return o1 instanceof CacheableWhitelist == o2 instanceof CacheableWhitelist ? 0 : o1 instanceof CacheableWhitelist ? -1 : 1;
```

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/CacheableWhitelist.java
```diff
@@ -0,0 +1,9 @@
+package org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists;
+
+/**
+ * Marker interface indicating results of permission checks do not change unless the list of entries is modified
+ *  and depend ONLY on the Field/Method/Constructor actually being invoked.
```
jglick: …and not the arguments to a method or constructor. To be explicit.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -122,77 +182,168 @@ public ProxyWhitelist(Whitelist... delegates) {
...
+                if (b == Boolean.FALSE && delegate instanceof CacheableWhitelist) {
+                    // Skip CacheableWhitelists if we already know none of them permit the method
+                    continue;
+                }
                 if (delegate.permitsConstructor(constructor, args)) {
+                    permittedCache.put(sigString, delegate instanceof CacheableWhitelist); // If we get a hit from non-Cacheable whitelists, none of the CacheableWhitelists permitted it
                     return true;
                 }
             }
+            permittedCache.put(sigString, Boolean.FALSE);
```
jglick: BTW can use autoboxing and write as simply `false`

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -666,7 +667,7 @@ public synchronized RejectedAccessException accessRejected(@Nonnull RejectedAcce
     }
 
     @Restricted(NoExternalUse.class) // implementation
-    @Extension public static final class ApprovedWhitelist extends ProxyWhitelist {
+    @Extension public static final class ApprovedWhitelist extends ProxyWhitelist implements CacheableWhitelist {
```
jglick: No. This includes `AclAwareWhitelist`, remember.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
```diff
@@ -66,6 +66,70 @@
...
+    }
+
+    /** Canonical name for a field access. */
+    public static String canonicalFieldString(@Nonnull Field field) {
+        return getName(field.getDeclaringClass()) + ' ' + field.getName();
+    }
+
+    /** Canonical name for a method call. */
+    public static String canonicalMethodString(@Nonnull Method method) {
+        return joinWithSpaces(new StringBuilder(getName(method.getDeclaringClass())).append(' ').append(method.getName()), argumentTypes(method.getParameterTypes())).toString();
```
jglick: No, this is **not safe**. You must encode the defining `ClassLoader` somehow. Ditto for all the other key calculations.

jglick: Could be fixed by using a double-level `WeakHashMap<ClassLoader, Map<String, Boolean>>` I think, since the values would not refer to the keys.

jglick: or a `WeakHashMap<Class, String>` if you just omit the `Class.name` from the start of the `String` part

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
jglick: …and even if it did not, it would not be cacheable without some sort of reset flag, since its delegates are mutable. Basically no `ProxyWhitelist` should be a `CacheableWhitelist` IIUC.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
svanoort: @jglick This does not make sense -- the Cacheable whitelists are using fixed string signatures -- https://github.com/jenkinsci/script-security-plugin/blob/8a4fa975bc182e3a365db1f9d53b20c3c62015c9/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist#L94 -- those signatures do not include a ClassLoader, nor can they.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
svanoort: Oops, that's a legit bug -- that was not intentional and slipped through.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
svanoort: Basically: if the static whitelists don't generate a hit, then we inspect the dynamic ones which can look at ClassLoader, etc

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
svanoort: What, no love for clever (ab)use of operators?

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
svanoort: Done

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelist.java
```diff
@@ -52,7 +52,7 @@
 /**
  * Whitelist based on a static file.
  */
-public final class StaticWhitelist extends EnumeratingWhitelist {
+public final class StaticWhitelist extends EnumeratingWhitelist implements CacheableWhitelist {
```
jglick: Redundant; `EnumeratingWhitelist` was already so marked.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -27,6 +27,7 @@
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.AclAwareWhitelist;
+import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.CacheableWhitelist;
```
jglick: Unused import.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/CacheableWhitelist.java
```diff
@@ -0,0 +1,10 @@
+package org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists;
+
+/**
+ * Marker interface indicating results of permission checks do not change unless the list of entries is modified
+ *  and depend ONLY on the Field/Method/Constructor actually being invoked, irrespective of classloader.
+ *  That means you must be able to represent the WhiteList entry with a String signature of className + method/field signature.
```
jglick: Setting aside `BlanketWhitelist`, which is not actually used in production and so need not actually be marked, I think what you really mean is that this should be applied only to `EnumeratingWhitelist`s, or to `ProxyWhitelist`s which only ever set their delegate once, and to an `EnumeratingWhitelist`. And these latter could actually be fixed to use the otherwise unusual field-scoped `@Extension`:

```java
public class JenkinsWhitelist {
    @Extension
    public static final Whitelist INSTANCE;
    static {
        try {
            INSTANCE = StaticWhitelist.from(JenkinsWhitelist.class.getResource("jenkins-whitelist"));
        } catch (IOException x) {
            throw new ExceptionInInitializerError(x);
        }
    }
```

Furthermore, change `EnumeratingWhitelist` to cache its signatures on demand, rather than repeatedly calling the methods. Since its check methods are `final`, we can then be assured that it is really safe to cache calls to it. Then delete `CacheableWhitelist` (so there is no chance of someone misusing it, as you accidentally did earlier in this PR! too easy to miss such subtleties) and simply check for `EnumeratingWhitelist` from `ProxyWhitelist`. Which we already do (including sorting the `EnumeratingWhitelist`s at the top of the list, since it is really a set)—so perhaps most of the rest of this PR could be discarded as well, assuming the only real problem at runtime was that we had some things like `JenkinsWhitelist` that were opaque to `ProxyWhitelist`’s optimizations for a fairly silly reason.

jglick:
Actually even the change to `JenkinsWhitelist` etc. might be unnecessary—`ProxyWhitelist` is already handling `ProxyWhitelist` delegates. I suspect it would suffice to better implement `equals`/`hashCode` on `EnumeratingWhitelist.Signature` subclasses (or just store the `toString` equivalents), and do some local optimizations (no API changes) to `ProxyWhitelist`: keep `methodSignatures` etc. as `Set`s, look up entries in there specially in `permitsMethod` etc. overrides as an initial step rather than delegating to a phony initial `EnumeratingWhitelist`, and change the locking semantics to optimize for reads. I see some of those basic notions in #180 and #181 but with other refactorings inherited from this one which are probably deadweight—would be more easily done from scratch I suspect.

(I always had a vague notion to do optimizations along these lines one day when they were needed but, hey, without a baseline performance measurement it seemed unwise.)

svanoort:
@jglick What I hadn't done with the HashSet based implementation in #181 yet was deal with the wildcards for identifiers in Signatures (see `EnumeratingWhitelist#is(String, String)` ).   That means we'd need to pull out the wildcards and iterate over them separately if the HashSet exact-match doesn't score a hit.  In practice I'm not *sure* if we're using wildcards anywhere yet but the code path nominally needs to stay supported.

We *could* create some new implementation like `OptimalStaticWhitelist` or similar that does this and have `ProxyWhiteList` use it internally.  We'd need some more extensive testing of the Wildcard cases, of course.

But from scalability testing https://github.com/jenkinsci/script-security-plugin/pull/180 gives us a lot of the benefits once the cache gets properly warmed up (which will take a bit) and is *much* easier to reason about.   

How much does it help?  Well with my simple test job running at 40x concurrency, we went from the lock on `delegates` being just as bad as XStream synchronization to it being almost irrelevant once the initial cache warming is done.

After spinning out the possibilities a bit more, I think we should just grab that approach and run with it -- no sense letting the perfect be the enemy of the good, and that allows time to attack the XStream bottleneck.

svanoort:
Closing in favor of #180 which is (sigh) a little simpler -- although this approach works that one's easier.

svanoort:
@jglick Now that #180 enables pre-caching of all permitted methods, I do not think we need this anymore -- it is performance-superior while being less complex and risky than #181.

jglick:
> we'd need to pull out the wildcards and iterate over them separately if the HashSet exact-match doesn't score a hit

Not necessarily. Given a method call like `ArrayList.ensureCapacity`, if the signature for that method was not in the set, you could check whether `ArrayList.*` was in the set; failing both of those, you go to the slow route of checking non-`EnumeratedWhitelist`s one by one.

> I'm not sure if we're using wildcards anywhere yet

I do not think so.

> nominally needs to stay supported

Needs to continue to _work_, for compatibility reasons, but need not be _fast_.

PR #180, opened 2018-02-05
Optimize EnumeratingWhitelist.permits with a simple cache

svanoort:
We've identified that Whitelist `permitsX` checks are a significant bottleneck to pipelines running CPS code.  Particularly because we hold a global lock on the top-level `ProxyWhitelist.delegates` while scanning the `ArrayList` of possible Whitelists.  Most of the time is spent iterating through lists of possible matches in the aggregated `EnumeratingWhitelist` (combines all the others into one).   This limits how much Pipeline code can be run concurrently, and is highly visible when looking at lock contention in a profiler.

This adds caching to `EnumeratingWhitelist`s and pre-populates the consolidated list used in the top-level ProxyWhitelist, which contains the union of all statically permitted signatures. 

In testing, lock contention becomes zero now that we're using a `ReentrantReadWriteLock` for the delegates.

Tests:

* Verified that the permitsX checks get solid coverage from the SandboxInterceptorTest in #181  (caught some bugs specific to that implementation) -> this shouldn't need any additional testing because existing tests cover it well.
* [x] Needs a testcase for the isWildcard method on Signatures
* [x] Needs a testcase for the ProxyWhitelist triggering/untriggering precaching of EnumeratingWhitelists
* [x] Added explicit testcases for permits & canonical signature creation for Whitelist lookups.


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/GenericWhitelist.java
```diff
@@ -33,7 +33,7 @@
  * Includes entries useful for general kinds of scripts.
  */
 @Restricted(NoExternalUse.class)
-@Extension public final class GenericWhitelist extends ProxyWhitelist {
+@Extension public final class GenericWhitelist extends StaticWhitelist {
```
svanoort: Safe because it's `NoExternalUse`

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/JenkinsWhitelist.java
```diff
@@ -33,7 +33,7 @@
  * Includes entries useful for scripts accessing the Jenkins API, such as model objects.
  */
 @Restricted(NoExternalUse.class)
-@Extension public final class JenkinsWhitelist extends ProxyWhitelist {
+@Extension public final class JenkinsWhitelist extends StaticWhitelist {
```
svanoort: Safe because it's `NoExternalUse`

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
```diff
@@ -66,6 +68,79 @@
 
     public abstract boolean permitsStaticFieldSet(@Nonnull Field field, @CheckForNull Object value);
 
+    // Utility methods for creating canonical string representations of the signature
+    public static final StringBuilder joinWithSpaces(StringBuilder b, String[] types) {
```
svanoort: `joinWithSpaces` and `getName` (both variants) and `argumentTypes` are all just relocated from EnumeratingWhitelist unchanged (allowing them to be reused here).

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelist.java
```diff
@@ -52,14 +52,22 @@
...
-public final class StaticWhitelist extends EnumeratingWhitelist {
+public class StaticWhitelist extends EnumeratingWhitelist {
 
     final List<MethodSignature> methodSignatures = new ArrayList<MethodSignature>();
     final List<NewSignature> newSignatures = new ArrayList<NewSignature>();
     final List<MethodSignature> staticMethodSignatures = new ArrayList<MethodSignature>();
     final List<FieldSignature> fieldSignatures = new ArrayList<FieldSignature>();
     final List<FieldSignature> staticFieldSignatures = new ArrayList<FieldSignature>();
 
+    public StaticWhitelist(StaticWhitelist list) {
```
jglick: Probably easier and clearer to take a `URL` argument, as in `from`.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -50,90 +50,106 @@
...
+                output = true;
+                break;
             }
         }
-        return false;
+        permittedCache.put(key, output);
+        return output;
     }
 
-    public static @Nonnull String getName(@Nonnull Class<?> c) {
```
jglick: Incompatible. You cannot move a `static` method to a “superclass”; only javac follows that lookup chain, not the JVM AFAIK.

Anyway `EnumeratingWhitelist` was the right place for these to begin with I think.

svanoort: Yeah, yeah -- I need to do a bit of cleanup on the PR before it's ready for full review.  That's why it's WIP - I wanted to spin out a few ideas before finalizing.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
svanoort: Note:  I'm retaining these here with a shim in `EnumeratingWhitelist` because these much more logically belong here -- providing a canonical identifier.  Also useful if we do another round of optimization or add another implementation, because these Strings make logical caching keys.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -113,6 +117,10 @@ public final void reset(Collection<? extends Whitelist> delegates) {
             for (ProxyWhitelist pw : wrappers.keySet()) {
                 pw.reset();
             }
+            if (this.wrappers.isEmpty()) {  // Top-level ProxyWhitelist should be the only cache
+                // Top-level ProxyWhitelist should precache the statically permitted signatures
+                ((EnumeratingWhitelist)(this.delegates.get(0))).precache();
+            }
```
svanoort: **TODO** we need a trivial test that the ProxyWhitelist caching actually correctly caches where it should.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -287,6 +344,11 @@ boolean matches(Field f) {
                 return false;
             }
         }
+
+        @Override
+        public boolean isWildcard() {
+            return "*".equals(field);
```
svanoort: TODO: needs a trivial testcase

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
```diff
@@ -66,6 +69,66 @@
...
+    }
+
+    /** Canonical name for a constructor call. */
+    @Restricted(NoExternalUse.class)
+    public static String canonicalConstructorString(@Nonnull Constructor cons) {
+        return joinWithSpaces(new StringBuilder(EnumeratingWhitelist.getName(cons.getDeclaringClass())), argumentTypes(cons.getParameterTypes())).toString();
+    }
+
+    @Restricted(NoExternalUse.class)
+    public static String canonicalMethodSig(@Nonnull Method method) {
```
svanoort: All of the `canonicalXSig` methods need to generate results matching the output of `toString` for the corresponding `EnumeratingWhiteList.Signature` implementation.  Verified by test: https://github.com/jenkinsci/script-security-plugin/pull/180/files#diff-4ff0624894ffcd88f5f79bcb89bbc036R104

svanoort: Because these match the Signature toStrings, we can use either form as a key -- and can easily generate a canonical String for a permission check based on the Field/Method/etc. 

Note: we could go and rip out a lot of the xxSignature methods etc except that they're published interfaces.

jglick:
FTR: intended to supersede both #179 & #181.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -50,69 +54,127 @@
...
+            for (Signature s : list) {
+                if (!s.isWildcard()) { // Cache entries for wildcard signatures will never be accessed and just waste space
+                    permittedCache.put(s.toString(), Boolean.TRUE);
+                }
+            }
+        }
+    }
+
+    /** Prepopulates the "permitted" cache. */
+    public void precache() {
```
jglick: Should be `final`. Or better, not `public` at all, since this does not look like something you should ever call from outside. At least, it neglects to call `permittedCache.clear()` before it starts, so it would leave behind stale cache entries if you were to modify your implementation and then call `precache` yourself.

Whatever the behavior, this needs to be specified in class Javadoc, since IIUC we are changing the expected semantics of this class: you may no longer dynamically return different values from `xxxSignatures` methods.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -50,69 +54,127 @@
...
+        cacheSignatureList((List)methodSignatures(), (List)(newSignatures()),
```
jglick: Unchecked warnings here. Use `@SafeVarargs`.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -50,69 +54,127 @@
...
+                           (List)(staticMethodSignatures()), (List)(fieldSignatures()),
+                           (List)(staticFieldSignatures()));
+    }
+
+    /** Frees up memory used for the cache. */
+    void clearCache() {
+        this.permittedCache.clear();
+        this.permittedCache = new ConcurrentHashMap<String, Boolean>();
```
jglick: What is the point of having _both_ these statements?

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -50,69 +54,127 @@
...
         for (MethodSignature s : methodSignatures()) {
             if (s.matches(method)) {
-                return true;
+                output = true;
+                break;
             }
         }
-        return false;
+        permittedCache.put(key, output);
+        return output;
```
jglick: This makes no sense to me. If we already precached all the signatures, why are we even iterating them again now? If it is not in the cache, just `return false` right away. Then we could delete the `matches` methods too, and make `permittedCache` be a `Set<String>`.

I guess you are doing this just to handle wildcard signatures? But as I noted in #179, that could be more easily (and efficiently) handled by precaching wildcards (i.e., deleting `isWildcard` and its call in `cacheSignatureList`) and having `permitsMethod` just do something like

```java
return permittedCache.contains(Whitelist.canonicalMethodSig(method)) ||
    permittedCache.contains(Whitelist.canonicalMethodSigStar(method));
```

where `canonicalMethodSigStar` just adds `*` rather than `method.getName()`. Analogously for `permitsField`.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -113,6 +138,12 @@ public final void reset(Collection<? extends Whitelist> delegates) {
...
+        } finally {
+            writer.unlock();
```
jglick: Wow crazy that this is not `AutoCloseable`.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -92,11 +115,13 @@ public final void reset(Collection<? extends Whitelist> delegates) {
                     staticMethodSignatures.addAll(ew.staticMethodSignatures());
                     fieldSignatures.addAll(ew.fieldSignatures());
                     staticFieldSignatures.addAll(ew.staticFieldSignatures());
+                    ew.clearCache();
```
jglick: I do not actually think you need to do this. Just set up the cache in the `EnumeratingWhitelist` initializer and never touch it again.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -113,6 +138,12 @@ public final void reset(Collection<? extends Whitelist> delegates) {
             for (ProxyWhitelist pw : wrappers.keySet()) {
                 pw.reset();
             }
+            if (this.wrappers.isEmpty()) {  // Top-level ProxyWhitelist should be the only cache
+                // Top-level ProxyWhitelist should precache the statically permitted signatures
+                ((EnumeratingWhitelist)(this.delegates.get(0))).precache();
```
jglick: More easily, make the cache be done in `EnumeratingWhitelist.<init>` (as above), and defer the creation of this until now:

```java
this.delegates.add(0, new EnumeratingWhitelist() {…});
```

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
svanoort: @jglick HashMaps don't shrink once expanded and there's no `arrayList.trimToSize()` equivalent.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
svanoort: Enumerating whitelists will likely be created before a ProxyWhitelist wrapping them -- no sense in forcing full cache initialization when it'll just get blown away again.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
svanoort: See answer below -- we're avoiding redundant cache generation.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
svanoort: Yes, it's to deal with wildcards.  We could indeed double-probe using a wildcard signature -- more elegant, but this implementation is already massively improved vs. the previous.

Can always be addressed later if needed -- I doubt the performance difference will matter much, since we're already caching FALSE.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
jglick: Yeah, though cache generation is cheap and just happens a few times during startup.

At any rate, could be done easily if desired by just having the cache field be nullable—initialize on demand.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
jglick: Have the annotation now, so should be able to remove the casts and thus be fully checkable.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -66,21 +66,24 @@ private void cacheSignatureList(List<Signature> ...sigs) {
...
+            this.permittedCache.clear();  // No sense calling clearCache
+        }
         cacheSignatureList((List)methodSignatures(), (List)(newSignatures()),
                            (List)(staticMethodSignatures()), (List)(fieldSignatures()),
                            (List)(staticFieldSignatures()));
     }
 
-    /** Frees up memory used for the cache. */
-    void clearCache() {
+    /** Frees up nearly all memory used for the cache.  MUST BE CALLED if you change the result of the xxSignatures() methods. */
```
jglick: But of course no known implementations do so, and as this is package-private no outside implementations could start doing so.

Again I think it would be wise to update the class Javadoc to emphasize that you are not allowed to dynamically change the signatures. If you need that for some weird reason, you would need to extend `Whitelist` directly, or (more likely) extend `ProxyWhitelist` and change the delegates.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
```diff
@@ -35,7 +35,10 @@
 import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.GroovySandbox;
+import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.EnumeratingWhitelist;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.ProxyWhitelist;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
```
jglick: unused imports

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
jglick: > I doubt the performance difference will matter much

Agreed, was suggesting this more to simplify the implementation. I care more about that in this code area than others because of the possible security implications of mistakes.

svanoort:
@jglick I don't have my original baseline handy unfortunately but we went from massive lock contention with 40x concurrent builds to... well, zero (all ReadLocks only).  The amount of time lost due to blocked threads was _not_ trivial.

PR #181, opened 2018-02-06
Optimize StaticWhitelist checks using a Hashset of canonical signature strings

svanoort:
Alternative to #179 that instead uses a HashSet of signatures constructed using canonical forms. 

Pros: 
* Precomputes the full lookup table, so no contention in ProxyWhitelist while the cache of static whitelists is being constructed. 

Cons:
* Requires rehashing the full signature list when the ProxyWhitelist is modified (might be an issue if you approve a LOT of signatures one after another). 
* Sensitive to canonical String format (has test coverage)
* Somewhat dirtier changes to structures

# TODO

* [ ] Support for the wildcard signatures (separate list with some sort of keying used)
* [ ] Might want to simply make a new implementation that is used by ProxyWhiteList internally and does *NOT* inherit from EnumeratingWhitelist


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -50,9 +50,7 @@
 
     protected abstract List<FieldSignature> staticFieldSignatures();
 
-    // TODO should precompute hash sets of signatures, assuming we document that the signatures may not change over the lifetime of the whitelist (or pass them in the constructor)
-
-    @Override public final boolean permitsMethod(Method method, Object receiver, Object[] args) {
+    @Override public boolean permitsMethod(Method method, Object receiver, Object[] args) {
```
jglick: Making these non`final` smells like a bad approach. If nothing else, blocks future optimizations.

svanoort: It's not my favorite thing definitely, but this is what permits us to use the HashSet lookup. 

Unfortunately @jglick our hands are somewhat tied because `EnumeratingWhitelist` insists on returning a `List<SomeSignatureType>` for the various `blahSignatures` methods.  IMO that was a mistake -- if they took a Collection then we could internally rework the implementation perhaps to do something clever with `Set`s and then fall back to binary search or iteration, but that ship has sailed.

PR #182, opened 2018-02-08
One more optimization and a cleanup change and doc change requested

svanoort:
@jglick Some leftovers from #180 - this way I don't have to worry too much about that blocking the original PR if it introduces some contentious point.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/AclAwareWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/AclAwareWhitelist.java
@@ -50,6 +50,9 @@ public class AclAwareWhitelist extends Whitelist {
      */
     public AclAwareWhitelist(Whitelist unrestricted, Whitelist restricted) {
         this.unrestricted = unrestricted;
+        if (this.unrestricted instanceof EnumeratingWhitelist) {
+            ((EnumeratingWhitelist) this.unrestricted).precache();
+        }
         this.restricted = restricted;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/AclAwareWhitelist.java
```diff
@@ -50,6 +50,9 @@
      */
     public AclAwareWhitelist(Whitelist unrestricted, Whitelist restricted) {
         this.unrestricted = unrestricted;
+        if (this.unrestricted instanceof EnumeratingWhitelist) {
+            ((EnumeratingWhitelist) this.unrestricted).precache();
```
jglick: OK, though as before it would be more modular to just have `EnumeratingWhitelist` set up its cache on its own the first time it is used.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -39,7 +39,11 @@
 import javax.annotation.Nonnull;
 
 /**
- * A whitelist based on listing signatures and searching them.
+ * A whitelist based on listing signatures and searching them. Lists of signatures should not change
+ * from invocation to invocation.
+ *
```
jglick:
```html
<p>
```

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -39,7 +39,11 @@
...
+ * If that's a need it is better to directly extend {@link Whitelist} and roll a custom implementation OR
+ *  extend ProxyWhitelist and add some custom delegates.
```
jglick: `@link`

PR #184, opened 2018-02-26
Replace "unclassified ..." error messages with something more helpful

abayer:
Not a *ton* more helpful, but I think it's better than just "unclassified ...".

I've seen enough cases of people (including myself) getting confused by what "unclassified ..." means that it feels like explaining that it means we couldn't find that particular field/constructor/method/etc is better than being vague. That said, I know it's not *quite* right in all cases, so I need to think on it more.

cc @reviewbybees


oleg-nenashev:
I do not think it will help much, but why not?

oleg-nenashev:
argh, sorry for clicking wrong button

PR #190, opened 2018-03-12
Switch to groovy-sandbox 1.17 release.

abayer:
cc @reviewbybees


abayer:
Will kick another rebuild once `groovy-sandbox` 1.17 actually shows up in Maven Central...

PR #191, opened 2018-03-15
[JENKINS-44924] Whitelist DefaultGroovyMethods for sort, unique, etc

abayer:
[JENKINS-44924](https://issues.jenkins-ci.org/browse/JENKINS-44924)

Downstream of https://github.com/cloudbees/groovy-cps/pull/86 (for all intents and purposes - it's not *literally* downstream, but...)

cc @reviewbybees


jglick:
> for all intents and purposes

Not really, you could use these methods from non-Pipeline scripts.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -366,9 +367,13 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods leftShift java.uti
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods max java.lang.Iterable groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods max java.lang.Object[] groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods max java.util.Collection groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods max java.util.Iterator groovy.lang.Closure
```
jglick: > If the closure has two parameters it is used like a traditional Comparator. I.e. it should compare its two parameters for order, returning […]. Otherwise, the Closure is assumed to take a single parameter and return a Comparable (typically an Integer) which is then used for further comparison.

And this is why the supreme being created static typing. /me sighs

PR #194, opened 2018-03-27
Add (Collection|List).remove methods to whitelist

mkobit:


mkobit:
Thanks @abayer for getting to these so fast! I know lots of PRs could be annoying, but I'm trying to keep them small enough if there are any complications or security considerations!

abayer:
@mkobit Not a problem in the least - you got a little lucky in that I was here mucking about for #192 yesterday and today anyway. =) FYI, I'm doing a release as we speak of 1.43.

PR #196, opened 2018-04-04
Add the ability to use minus method with strings and def

raul-arabaolaza:
I think is a useful method, so you can use code like

```
amethod(url) {
    println(url - "file://")
}
```


abayer:
LGTM - thanks!

PR #199, opened 2018-04-17
Allow calling Closure elements of Maps as methods

rudolfwalter:
The sandbox rejects valid Groovy code when trying to call a Closure element of a Map as if it were one of the Map's methods. See [JENKINS-50843](https://issues.jenkins-ci.org/browse/JENKINS-50843).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
 
 import com.google.common.collect.ImmutableSet;
+import groovy.lang.Closure;
 import groovy.lang.GroovyRuntimeException;
 import groovy.lang.MetaMethod;
 import groovy.lang.MissingMethodException;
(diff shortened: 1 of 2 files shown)


svanoort:
@rudolfwalter I think it would be helpful to have Closures taking parameters in the test, yeah. 

Ignore the comment about functions.

rudolfwalter:
@svanoort done :)

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -114,6 +116,14 @@
                 throw StaticWhitelist.rejectStaticMethod(foundDgmMethod);
             }
 
+            // allow calling Closure elements of Maps as methods
+            if (receiver instanceof Map) {
+                Object element = onMethodCall(invoker, receiver, "get", method);
```
jglick: Not technically correct to delegate to `onMethodCall` here: if the whitelist did not pass this particular `Map.get(Object)` call but the receiver _was_ allowed to work in some other way like the `invokeMethod` below, then we would be breaking something that worked before. OTOH `Map.get(Object)` at the interface level is whitelisted, so using the standard whitelist there is no way this would not pass.

Just another example of the problem noted in https://github.com/jenkinsci/groovy-sandbox/issues/7#issuecomment-39252544.

rudolfwalter:
@abayer Kind reminder that this needs to be merged.

svanoort:
@rudolfwalter Don't worry, I've got this.

PR #203, opened 2018-05-11
Whitelist: add jsonbuilder

avindra:
This should allow code such as:

```groovy
new groovy.json.JsonBuilder([
    some: "value",
    iwant: "json",
    pls: "jenkins"
]).toString() 
```

Which is a fairly common use case, e.g., in posting data to a url (for notifications) in post-build actions.


jglick:
I am not sure this is safe. Do not merge without careful analysis. In general, use of Groovy’s JSON tooling from Pipeline Script is discouraged.

avindra:
Found a way around this, ignore

PR #209, opened 2018-07-31
Whitelist a useful subset of the Java 8 time API

TobiX:
These are quite useful to calculate daily timestamps...

(Currently only Local* and no Zoned*, since I didn't require anything else...)

PS: It looks like the code understands wildcards in signatures, which would have been quite useful here, but the test suite denies me their use... Is this deliberate?


svanoort:
Thanks for the contribution!  Merging now

PR #210, opened 2018-08-08
[FIXED JENKINS-51332] Whitelist Calendar.MONDAY/JANUARY et al

abayer:
[JENKINS-51332](https://issues.jenkins-ci.org/browse/JENKINS-51332)


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -114,20 +114,39 @@ new java.text.SimpleDateFormat java.lang.String
...
 staticField java.util.Calendar MONTH
+staticField java.util.Calendar NOVEMBER
+staticField java.util.Calendar OCTOBER
+staticField java.util.Calendar SATURDAY
 staticField java.util.Calendar SECOND
+staticField java.util.Calendar SEPTEMBER
+staticField java.util.Calendar SUNDAY
+staticField java.util.Calendar THURSDAY
+staticField java.util.Calendar TUESDAY
+staticField java.util.Calendar WEDNESDAY
```
dwnusbaum: No `UNDECIMBER`? Half-joking, but maybe it makes sense to add it for consistency?

dwnusbaum:
Two other fields that seem like they might be used often enough to add are `Calendar.AM` and `Calendar.PM`, especially since we already whitelisted `Calendar.AM_PM`.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
abayer: wat. =)

abayer:
Good call on `AM` and `PM` - added.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
dwnusbaum: https://stackoverflow.com/questions/44716517/what-does-calendar-undecimber-do 😄

abayer: I'm still saying wat and running away from that. =)

PR #211, opened 2018-08-08
[FIXED JENKINS-50906] Allow foo() for Script binding closure variables

abayer:
[JENKINS-50906](https://issues.jenkins-ci.org/browse/JENKINS-50906)

Interestingly, this doesn't show up in Pipeline, just in vanilla sandboxed Groovy scripts.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
@@ -1089,4 +1089,10 @@ public class SandboxInterceptorTest {
         assertEvaluate(new GenericWhitelist(), "a=hello,b=10", "def m = [ f: {a,b -> return \"a=${a},b=${b}\"} ]; m.f('hello',10)");
         assertEvaluate(new GenericWhitelist(), 2, "def m = [ f: {it.size()} ]; m.f(foo:0, bar:1)");
     }
+
+    @Issue("JENKINS-50906")
+    @Test
+    public void scriptBindingClosureVariableCall() throws Exception {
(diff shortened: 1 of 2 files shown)


svanoort:
I'll trust review by @dwnusbaum here -- if he thinks it's good, ship it :)

dwnusbaum:
AFAICT this just adds support for some Groovy syntactic sugar where we already support the no-sugar equivalent, so it looks good to me.

jglick:
CI problems, trigger a rebuild.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -1089,4 +1089,10 @@ public void callClosureElementOfMapAsMethod() throws Exception {
         assertEvaluate(new GenericWhitelist(), "a=hello,b=10", "def m = [ f: {a,b -> return \"a=${a},b=${b}\"} ]; m.f('hello',10)");
         assertEvaluate(new GenericWhitelist(), 2, "def m = [ f: {it.size()} ]; m.f(foo:0, bar:1)");
     }
+
+    @Issue("JENKINS-50906")
+    @Test
+    public void scriptBindingClosureVariableCall() throws Exception {
+        assertEvaluate(new GenericWhitelist(), true, "def func = { 1 }; this.func2 = { 1 }; return func() == func2();\n");
```
jglick: A single variant is not enough to convince readers that the use of `(… "call", args)` works correctly as varargs. I would expect to see closures with zero, one, and two or more parameters, as well as implicit `it`.

abayer: Fair enough - will do.

PR #214, opened 2018-08-21
JENKINS-53156: Whitelist Map.remove()

haridsv:
This fixes: [JENKINS-53156](https://issues.jenkins-ci.org/browse/JENKINS-53156)


haridsv:
I see a test failure and it looks like they need to be sorted? Let me reorder it and see if the issue goes away.

PR #218, opened 2018-08-29
Use lock for manipulation with wrapper too.

lvotypko:
I hit the issue [JENKINS-41797](https://issues.jenkins-ci.org/browse/JENKINS-41797) again.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
@@ -150,8 +150,12 @@ public class ProxyWhitelist extends Whitelist {
 
     private void addWrapper(ProxyWhitelist pwl){
         lock.writeLock().lock();
-        wrappers.put(pwl, null);
-        lock.writeLock().unlock();
+        try {
+            wrappers.put(pwl, null);
+        }
+        finally {
+            lock.writeLock().unlock();
+        }
     }
 
 
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
@@ -65,6 +65,7 @@ public class ProxyWhitelist extends Whitelist {
     private final List<EnumeratingWhitelist.FieldSignature> staticFieldSignatures = new ArrayList<EnumeratingWhitelist.FieldSignature>();
 
     /** anything wrapping us, so that we can propagate {@link #reset} calls up the chain */
+    @GuardedBy("lock")
     private final Map<ProxyWhitelist,Void> wrappers = new WeakHashMap<ProxyWhitelist,Void>();
 
     // TODO Consider StampedLock when we switch to Java8 for better performance - https://dzone.com/articles/a-look-at-stampedlock
@@ -118,7 +119,7 @@ public class ProxyWhitelist extends Whitelist {
                     ew.clearCache();
                 } else if (delegate instanceof ProxyWhitelist) {
                     ProxyWhitelist pw = (ProxyWhitelist) delegate;
-                    pw.wrappers.put(this, null);
+                    pw.addWrapper(this);
                     for (Whitelist subdelegate : pw.delegates) {
                         if (subdelegate instanceof EnumeratingWhitelist) {
                             ((EnumeratingWhitelist) subdelegate).clearCache();  // We only care about top-level cache
@@ -146,6 +147,13 @@ public class ProxyWhitelist extends Whitelist {
             writer.unlock();
         }
     }
+
+    private void addWrapper(ProxyWhitelist pwl){
+        lock.writeLock().lock();
+        wrappers.put(pwl, null);
+        lock.writeLock().unlock();
+    }
+
 
     public ProxyWhitelist(Whitelist... delegates) {
         this(Arrays.asList(delegates));


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -147,6 +148,13 @@ public final void reset(Collection<? extends Whitelist> delegates) {
         }
     }
 
+    private void addWrapper(ProxyWhitelist pwl){
+        lock.writeLock().lock();
```
dwnusbaum: Could we get deadlock now in some cases? `#reset` holds a lock for the outer ProxyWhitelist, and when this is called it will acquire the lock for the delegate ProxyWhitelist. Do we know that the locks will always be acquired in the same order? I would be surprised if there are cyclic references between instances of ProxyWhitelist, so my guess is that it should be fine, but it would be good for someone to confirm.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -147,6 +148,13 @@ public final void reset(Collection<? extends Whitelist> delegates) {
...
+        wrappers.put(pwl, null);
+        lock.writeLock().unlock();
```
dwnusbaum: Even if we don't expect `WeakHashMap#put` to throw an exception, I think it would be a good idea to use a `try`/`finally` block for the lock.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
svanoort: Cyclic references shouldn't be possible, but this makes me nervous too -- although for different reasons: I wonder what happens if we have two different `reset` operations running in parallel, starting from different ProxyWhitelists.  

Note also that read and write locks are independent - you can't "upgrade" a readlock to a writelock.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
svanoort: I agree -- this is a clear 🐛  for me, because an unexpected exception can prevent us ever unlocking.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
svanoort: AFAICT we actually ought to obtain one of the delegate's locks  when reading its fields, meaning this whole block theoretically should be synchronized (probably with a WriteLock for safety, since it also implies a readlock).

svanoort: Theoretically the lock orders should be consistent, since we're climbing up the tree of wrapping whitelists, but I worry about the fact that we also call reset on wrapper ProxyWhitelists further down (and have read calls coming in as well).   WDYT @jglick ?

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
lvotypko: Agree, I will add it.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
lvotypko: @dwnusbaum it makes me worry too, because I am not sure this is impossible too. I will think about writing at least some test for this, but honestly I wanted to write it for whole reset method and it is not such easy task. I was able to hit my issue with test but not in 100%, usually I hit it in 2 cases from 5. Increasing the number of loops did not make it more successful. The whole method made it less probably with increasing loops, since growing numbers of elements in map (since the wrappers are only added but never removed) increases time of execution somewhere else then where the issues is.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
jglick: [sigh](https://stackoverflow.com/a/17043350/12916)

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
jglick: > WDYT @jglick ?

It would take me a while to digest this and remember exactly how this code works. Any change here is pretty risky.

bitwiseman: @jglick @dwnusbaum 
Is this even needed?   We're already taking the write lock at the top of the calling method and there are not other places where this private field is accessed.

bitwiseman:
@jglick @dwnusbaum
Is this even needed? We're already taking the write lock at the top of the calling method and there are no other places where this private field is accessed.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
dwnusbaum: @bitwiseman The premise of the PR is that without a lock here, @lvotypko ran into a situation where `wrappers` was updated concurrently which caused an infinite loop. If we don't add a lock here, then this PR doesn't do anything. My best guess as to what happened is that two distinct whitelists delegated to the same whitelist, and reset was somehow called on them concurrently, but without a way to reproduce the issue it's hard to say.

jglick:
> Is this even needed?

As per https://github.com/jenkinsci/script-security-plugin/pull/218#discussion_r214408879 I do not know offhand. Whoever takes on review for this needs to analyze the existing logic closely.

bitwiseman:
Okay, let me restate this.  @dwnusbaum 
`wrappers` is `private`.  The only place it is accessed is inside `reset()` at the following three points:

https://github.com/jenkinsci/script-security-plugin/blob/9242cafd68b2572022ecbeefc6583acd8f4498ee/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java#L122

https://github.com/jenkinsci/script-security-plugin/blob/9242cafd68b2572022ecbeefc6583acd8f4498ee/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java#L139

https://github.com/jenkinsci/script-security-plugin/blob/9242cafd68b2572022ecbeefc6583acd8f4498ee/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java#L142

That whole method is covered by the writer lock already.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -118,7 +119,7 @@ public final void reset(Collection<? extends Whitelist> delegates) {
                     ew.clearCache();
                 } else if (delegate instanceof ProxyWhitelist) {
                     ProxyWhitelist pw = (ProxyWhitelist) delegate;
-                    pw.wrappers.put(this, null);
+                    pw.addWrapper(this);
```
bitwiseman:
```suggestion
                    pw.wrappers.put(this, null);
```

dwnusbaum:
@bitwiseman The lock in `reset` is for the outer whitelist. The new `addWrapper` method is called on the delegate whitelist, so it acquires the lock for the delegate whitelist, not the outer whitelist. There are multiple levels of locking involved with this change, hence my questions about lock order [here](https://github.com/jenkinsci/script-security-plugin/pull/218#discussion_r213741215).

bitwiseman:
@dwnusbaum 
Oh, I see it now. 🥇

lvotypko:
After an old discussion and good arguments that this pull request is not safe - or rather that it will fix one issue with concurrency but there is possibility to cause another one.  I can report that it has not happened again in our instance since I opened this ticket. So the question is does this have a big probability to happen (I think not because even reproduce it by test was not easy)? I propose to close pull request because this is clearly not often problem and fixing this can cause another problem (more and equal severe). Do you agree?

PR #219, opened 2018-08-30
Created a "glue-code" for casc plugin

odavid:
This PR enables the end user to configure method signature approvals using [casc plugin](https://github.com/jenkinsci/configuration-as-code-plugin)

See conversation in 
https://github.com/jenkinsci/configuration-as-code-plugin/pull/416#issuecomment-416973989

Since the casc plugin is still in beta, I've changed the pom to enable beta plugin dependencies. This property will/should be removed once the casc plugin 1.0 will be released

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,11 +67,5 @@
       <version>${version.configuration-as-code}</version>
       <optional>true</optional>
     </dependency>
-    <dependency>
-      <groupId>io.jenkins.configuration-as-code</groupId>
-      <artifactId>configuration-as-code-support</artifactId>
-      <version>${version.configuration-as-code}</version>
-      <optional>true</optional>
-    </dependency>
   </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -18,9 +18,7 @@
       <changelist>-SNAPSHOT</changelist>
       <jenkins.version>2.60.3</jenkins.version>
       <java.level>8</java.level>
-      <!-- TODO: Remove usBeta once the casc plugin will be released -->
-      <useBeta>true</useBeta>
-      <configuration-as-code.version>1.0</configuration-as-code.version>
+      <configuration-as-code.version>1.1</configuration-as-code.version>
     </properties>
     <licenses>
         <license>
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,10 @@
       <revision>1.47</revision>
       <changelist>-SNAPSHOT</changelist>
       <jenkins.version>2.7.3</jenkins.version>
-      <java.level>7</java.level>
+      <java.level>8</java.level>
+      <!-- TODO: Remove usBeta once the casc plugin will be released -->
+      <useBeta>true</useBeta>
+      <version.configuration-as-code>1.0-rc2</version.configuration-as-code>
     </properties>
     <licenses>
         <license>
@@ -57,6 +60,18 @@
           <artifactId>groovy</artifactId>
         </exclusion>
       </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>io.jenkins</groupId>
+      <artifactId>configuration-as-code</artifactId>
+      <version>${version.configuration-as-code}</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>io.jenkins.configuration-as-code</groupId>
+      <artifactId>configuration-as-code-support</artifactId>
+      <version>${version.configuration-as-code}</version>
+      <optional>true</optional>
     </dependency>
   </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -16,11 +16,11 @@
     <properties>
       <revision>1.47</revision>
       <changelist>-SNAPSHOT</changelist>
-      <jenkins.version>2.7.3</jenkins.version>
+      <jenkins.version>2.60.3</jenkins.version>
       <java.level>8</java.level>
       <!-- TODO: Remove usBeta once the casc plugin will be released -->
       <useBeta>true</useBeta>
-      <version.configuration-as-code>1.0</version.configuration-as-code>
+      <configuration-as-code.version>1.0</configuration-as-code.version>
     </properties>
     <licenses>
         <license>
@@ -64,7 +64,7 @@
     <dependency>
       <groupId>io.jenkins</groupId>
       <artifactId>configuration-as-code</artifactId>
-      <version>${version.configuration-as-code}</version>
+      <version>${configuration-as-code.version}</version>
       <optional>true</optional>
     </dependency>
   </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
       <java.level>8</java.level>
       <!-- TODO: Remove usBeta once the casc plugin will be released -->
       <useBeta>true</useBeta>
-      <version.configuration-as-code>1.0-rc2</version.configuration-as-code>
+      <version.configuration-as-code>1.0</version.configuration-as-code>
     </properties>
     <licenses>
         <license>
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/casc/ScriptApprovalConfigurator.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/casc/ScriptApprovalConfigurator.java
@@ -19,6 +19,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.logging.Logger;
 
 /**
  * @author <a href="mailto:ohad.david@gmail.com">Ohad David</a>
@@ -27,6 +28,7 @@ import java.util.Set;
 @Restricted(NoExternalUse.class)
 public class ScriptApprovalConfigurator extends BaseConfigurator<ScriptApproval> implements RootElementConfigurator<ScriptApproval> {
 
+    private static final Logger LOGGER = Logger.getLogger(ScriptApprovalConfigurator.class.getName());
 
     @Override
     public String getName() {
@@ -51,6 +53,7 @@ public class ScriptApprovalConfigurator extends BaseConfigurator<ScriptApproval>
         try {
             return target.approveSignature(signature);
         } catch (IOException e) {
+            LOGGER.warning("Could not approve signature: " + signature + " ScriptApproval threw IOException: " + e.getMessage());
             throw new UncheckedIOException(e);
         }
     }
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/casc/ScriptApprovalConfigurator.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/casc/ScriptApprovalConfigurator.java
@@ -1,0 +1,77 @@
+package org.jenkinsci.plugins.scriptsecurity.casc;
+
+import hudson.Extension;
+import io.jenkins.plugins.casc.Attribute;
+import io.jenkins.plugins.casc.BaseConfigurator;
+import io.jenkins.plugins.casc.ConfigurationContext;
+import io.jenkins.plugins.casc.RootElementConfigurator;
+import io.jenkins.plugins.casc.impl.attributes.MultivaluedAttribute;
+import io.jenkins.plugins.casc.model.CNode;
+import io.jenkins.plugins.casc.model.Mapping;
+import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+import javax.annotation.CheckForNull;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.Arrays;
+import java.util.Collection;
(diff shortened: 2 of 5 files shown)


odavid:
Hi,
Although I changed the [Jenkinsfile](https://github.com/jenkinsci/script-security-plugin/blob/ec56950fb25e98d17004cdec1d86c08f7cd63604/Jenkinsfile#L1) to use `2.7.3` Jenkins version, it seems the `2.60.2` are still running and failing because the `Configuration As Code` plugin does not support this version.

Is it hardcoded in Jenkins? Or am I missing something?

@ndeloof FYI^^

ndeloof:
@odavid `jenkinsVersions: [null` means build will first run with default jenkins version as set by pom.xml. You have to change both `java.leve=8` and `jenkins.version=2.60.3`.

pom.xml
```diff
@@ -56,5 +59,17 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>io.jenkins</groupId>
+      <artifactId>configuration-as-code</artifactId>
+      <version>${version.configuration-as-code}</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
```
ndeloof: you only need configuration-as-code plugin here, -support is useless

odavid: Fixed. was put by mistake...

odavid:
@ndeloof - I changed the pom and the Jenkinsfile to use 2.60.3, however it still tries to run 2.60.2 and fails. Is there anything else I need to do?

ndeloof:
build failure is caused by test failure https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fscript-security-plugin/detail/PR-219/6/tests
"
Caused by: java.io.IOException: Configuration as Code Plugin v1.0-rc2 failed to load.
- You must update Jenkins from v2.60.2 to v2.60.3 or later to run this plugin
"

and as you already noticed the Jenkinsfile was configured to run tests on 2.60.2

odavid:
But it is not. it is using 2.60.3 - https://github.com/jenkinsci/script-security-plugin/pull/219/commits/20a57e737820c2b826a023d2c5adbe4039d03203

ndeloof:
oh yes indeed. Strange then. I don't know how `buildPlugin` is implemented for this purpose, sounds it's failing to select adequate build environment

odavid:
Trying to remove versions at all. Seems to me a bug in the pipeline lib

odavid:
It seems the pipeline is hardcoded somewhere, otherwise I cannot explain this behavior. Not sure what I can do about it.

dwnusbaum:
@odavid Your changes to the Jenkinsfile are ignored because you don't have write access to the repository (it's a security feature).

odavid:
thanks @dwnusbaum - this is good to know 😄 - So, I am leaving the PR as is... Nothing I can do more than that...
Will be glad if it will be merged soon... Or at least when the CasC plugin will be released.

Thanks again,
Ohad

dwnusbaum:
@odavid No problem, thanks for the PR! Someone with write access to the repo can re-file your PR separately before merging to check that the CI build passes.

abayer:
I'd like to wait on this until JCasC hits 1.0.

odavid:
I understand, I will wait...
Thanks,
Ohad

src/test/resources/org/jenkinsci/plugins/scriptsecurity/casc/ScriptApprovalConfiguratorTest.yml
```diff
@@ -0,0 +1,5 @@
+scriptApproval:
+  approvedSignatures:
+    - method java.net.URI getHost
```
jglick: I do not think this is a good idea. We should not be encouraging administrators to create custom preapprovals. Genuinely safe methods can be added to the `generic-whitelist` (or similar) in this plugin and a new release cut. If it is not obvious whether a method is safe, do not whitelist it. Simplify your scripts to do processing in external processes.

ndeloof: do you mean the feature to let administrator add a signature to the whitelist _from the web UI_ is a bad idea ?

odavid: @jglick, @ndeloof - I understand that having custom approved signatures should not be encouraged, however I think that since this feature exists, it will be better for Administrators to be able to control this configuration within a source code repository and not using the UI. I think that any feature that can be configured through the UI, should have an ability to be configured through source code.
Best,
Ohad

odavid:
Hi There, Since the casc plugin was released, Is there any chance this PR will be merged?

ewelinawilkosz:
@odavid looking into that!

ewelinawilkosz:
@abayer poking you since you're the maintainer - any chance you could have a look and merge/comment? and is there any prediction regarding when it could be released (if approved)? no pressure, but since the question was raised by a number of users at Jenkins World I'd like to know what to tell them

I hear @jglick doubts but agree with @ndeloof - we're not providing new feature with jcasc, we want to make configurable whatever is possible to be configured via UI - unless you want to remove that fature from UI I can't see a reason why we can't support it with jcasc :)

src/test/resources/org/jenkinsci/plugins/scriptsecurity/casc/ScriptApprovalConfiguratorTest.yml
jhoblitt: Speaking as one who has been deploying jenkins in a test env with puppet for modifications, then pushing those changes into production for 4 years this is not a minor concern.  I have something like 30 signatures to approve -- which means having to run test pipelines after a production push to see if any new signatures are needed and the pain and misery have having to run a pipeline script + approve a signature in a loop 30 times for ever new sandbox env (typically several times per week and sometimes per day).

pom.xml
```diff
@@ -17,7 +17,10 @@
       <revision>1.47</revision>
       <changelist>-SNAPSHOT</changelist>
       <jenkins.version>2.7.3</jenkins.version>
```
oleg-nenashev: Should be changed to 2.60.3 then

pom.xml
```diff
@@ -17,7 +17,10 @@
...
-      <java.level>7</java.level>
+      <java.level>8</java.level>
+      <!-- TODO: Remove usBeta once the casc plugin will be released -->
+      <useBeta>true</useBeta>
+      <version.configuration-as-code>1.0</version.configuration-as-code>
```
oleg-nenashev: Usually Maven naming convention is `configuration-as-code.version`

pom.xml
```diff
@@ -58,5 +61,11 @@
...
     </dependency>
+    <dependency>
+      <groupId>io.jenkins</groupId>
+      <artifactId>configuration-as-code</artifactId>
+      <version>${version.configuration-as-code}</version>
+      <optional>true</optional>
+    </dependency>
   </dependencies>
-</project>
+</project>
```
oleg-nenashev: bad change

oleg-nenashev:
I agree with comments w.r.t the test case, but the rest of the code looks good

odavid:
Hey @oleg-nenashev - Not sure if you missed it, I've resolved all your comments. Is it going to be merged?

oleg-nenashev:
> Hey @oleg-nenashev - Not sure if you missed it, I've resolved all your comments. Is it going to be merged?

I do not watch this repo and I do not receive notifications unless explicitly pinged

pom.xml
```diff
@@ -16,8 +16,11 @@
     <properties>
       <revision>1.47</revision>
       <changelist>-SNAPSHOT</changelist>
-      <jenkins.version>2.7.3</jenkins.version>
-      <java.level>7</java.level>
+      <jenkins.version>2.60.3</jenkins.version>
+      <java.level>8</java.level>
+      <!-- TODO: Remove usBeta once the casc plugin will be released -->
```
oleg-nenashev: No longer required, you can switch to JCasC 1.1

src/main/java/org/jenkinsci/plugins/scriptsecurity/casc/ScriptApprovalConfigurator.java
```diff
@@ -0,0 +1,77 @@
...
+                .getter((ScriptApproval target) -> Arrays.asList(target.getApprovedSignatures()))
+                .setter((target, value) -> value.forEach(signature -> uncheckedApproveSignature(target, signature))));
+        return describe;
+    }
+
+    private static String[][] uncheckedApproveSignature(ScriptApproval target, String signature) {
+        try {
+            return target.approveSignature(signature);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
```
oleg-nenashev: This is a bad idea, it will still make JCasC to fail in unpredictable way.
I would rather expect an explicit failure to happen. If it is not possible, better to log a warning to the system log

odavid: @oleg-nenashev - Resolved both comments.

asquel:
@oleg-nenashev & @odavid : are you going to merge this stuff ?

src/main/java/org/jenkinsci/plugins/scriptsecurity/casc/ScriptApprovalConfigurator.java
oleg-nenashev: The current code still throws unchecked exceptions. CC @ndeloof who can comment about the viability of such approach

oleg-nenashev:
@asquel Only maintainers can do the merge. IIUC reviews from @svanoort @rsandell @dwnusbaum @abayer @olamy are needed here

src/main/java/org/jenkinsci/plugins/scriptsecurity/casc/ScriptApprovalConfigurator.java
ndeloof: I personally dislike checked exception, they prevented me to use method references in streams few times, and I use to just forward to caller, which at the end dump some stacktrace whatever happens.  I'd prefer java only define unchecked ones.

oleg-nenashev: There is Kotlin for it 🤷‍♂️

ndeloof: yes indeed. Maybe this is what we need to "Jenkins Bolt" to actually renew Jenkins developer experience

abayer:
I'm still not a fan of this - why does it need to be in `script-security` rather than in `configuration-as-code`? This is such a key plugin that I really don't want it to be dependent on anything more than absolutely necessary.

ewelinawilkosz:
I think it's general approach, that code specific for the plugin should be on plugin's site, if we want the plugin to be supported by configuration as code. It is either updating the implementation to follow conventions or providing dedicated configurator.

ndeloof:
the _other_ option is for this plugin to follow a conventional getter/(databound)setter approach to configuration, so no custom code is required for casc support.

- make ScriptApproval a GlobalConfiguration or something comparable casc can discover as a "configuration target"
- offer get/set methods to manage approved signatures

abayer:
@ndeloof I’d be willing to consider changes that make it easier for JCasC to interact with `script-security`, so long as there are no compatibility breaks.

ndeloof:
right. But such changes might be way more controversial vs an optional extra component

odavid:
@ndeloof, @abayer - As you know, I started this PR within JCasC repo... 
I see 2 alternatives in order for this functionality to be available.
* Move it back to JCasC
* Create a sub module within the `script-security-plugin` to serve the glue-code of JCasC.

Let me know your thoughts,
Ohad

jeff-knurek:
> @ndeloof, @abayer - As you know, I started this PR within JCasC repo...
I see 2 alternatives in order for this functionality to be available.
> - Move it back to JCasC
> - Create a sub module within the script-security-plugin to serve the glue-code of JCasC.

any update on a decision here, similar to @jhoblitt use case, this a significant issue for us

bblay:
Hi, there seems to be some agreement on moving back to jcasc.
Was this a viable pull request?
https://github.com/jenkinsci/configuration-as-code-plugin/issues/415

ndeloof:
@bblay you've misunderstood then. From my PoV JCasC should not own any plugin specific adapter.

odavid:
@ndeloof, @abayer,
I asked you for your thoughts here: https://github.com/jenkinsci/script-security-plugin/pull/219#issuecomment-429605819, but did not get replies yet, in the meantime, I see that 
#233 was created, so I assume it will replace it. Will you consider accepting #233 ?

If so, #233 should replace this PR, If not, please answer my questions regarding additional maven module/plugin that will be the glue code.

Best,
Ohad

abayer:
Sorry, I've been insanely busy and distracted. I'll try to look at #233 this week.

bblay:
In case it's helpful to others:
We worked around this problem by moving our function into a shared library, which needs no security approval.

jglick:
Superseded by #233 I think.

PR #220, opened 2018-08-30
Whitelist asImmutable java.util.Map.

valentinamanea:
There should be possible to make a map immutable in a Groovy script.


dwnusbaum:
Thanks for cleaning up the stray file as well @valentinamanea!

PR #222, opened 2018-09-04
Downgrade warning about pathologically slow load times

jglick:
Appears in Sentry logs for Evergreen. While something is very likely wrong in this circumstance, it is not necessarily a software issue, so it is not generally an actionable warning for developers.


batmat:
Can we expect a release with this soon?

I see the last release was in April by @abayer and there's been quite a few changes since then. I guess the more we wait, the higher the risk grows, given the growing diff chunk size grows since `1.44` :)?

Thanks!

dwnusbaum:
@batmat  I plan to release the plugin later today

PR #223, opened 2018-09-05
[FIXED JENKINS-53240] Check for variable in binding first.

abayer:
[JENKINS-53240](https://issues.jenkins-ci.org/browse/JENKINS-53420)

Logic here now is derived from `CpsScript#invokeMethod`.

Downstream validation of the fix coming in a PR to workflow-cps momentarily.


abayer:
(and yes, @jglick, we should incrementalize `script-security`, but right at the moment I'm just focused on the fix)

abayer:
@jglick I'll try - I'm honestly not sure how easy it'll be, since the issue is specifically with how we handle steps vs actual methods.

jglick:
I see.

BTW this regression is a prime example of the need for [JENKINS-45047](https://issues.jenkins-ci.org/browse/JENKINS-45047).

abayer:
@jglick Yes it definitely is.

jglick:
> we should incrementalize `script-security`

#224. Literally took two minutes FWIW.

jglick:
(Update parent, then `mvn incrementals:incrementalify`, if you have never done it before.)

PR #227, opened 2018-10-15
Add missing DefaultGroovyMethods minus/plus whitelist entries

TobiX:
This makes it easier to write idiomatic groovy code without whitelisting every list/set operation 😉

Theses should be all public plus/minus methods for current Groovy trunk


svanoort:
Thanks for the contribution!  Merging now, should go into the next release.

PR #232, opened 2018-11-25
Update SecureGroovyScript.java

sternr:
Making sure expensive log lines are calculated only when log level is enabled

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
@@ -221,7 +221,7 @@ public final class SecureGroovyScript extends AbstractDescribableImpl<SecureGroo
             ClassLoader encounteredLoader = klazz.getClassLoader();
             if (encounteredLoader != loader) {
                 it.remove();
-                if (LOGGER.isLoggable(Level.FINER)) {
+                if (LOGGER.isLoggable(Level.FINEST)) {
                   LOGGER.log(Level.FINEST, "ignoring {0} with loader {1}", new Object[] {klazz, /* do not hold from LogRecord */String.valueOf(encounteredLoader)});
                 }
             }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -219,7 +221,9 @@ private static void cleanUpGlobalClassValue(@Nonnull ClassLoader loader) throws
             ClassLoader encounteredLoader = klazz.getClassLoader();
             if (encounteredLoader != loader) {
                 it.remove();
-                LOGGER.log(Level.FINEST, "ignoring {0} with loader {1}", new Object[] {klazz, /* do not hold from LogRecord */String.valueOf(encounteredLoader)});
+                if (LOGGER.isLoggable(Level.FINEST)) {
+                  LOGGER.log(Level.FINEST, "ignoring {0} with loader {1}", new Object[] {klazz, /* do not hold from LogRecord */String.valueOf(encounteredLoader)});
+                }
             }
         }
         LOGGER.log(Level.FINE, "cleaning up {0} associated with {1}", new Object[] {toRemove.toString(), loader.toString()});
```
dwnusbaum: Any reason not to change this line and a few others that are similar? Did you change the lines you did based on profiling or something showing that too much time was spent in them (maybe because they occur in loops or something)?

sternr: Yes, I have a bunch of thread dumps showing String.valueOf(encounteredLoader) as a hot place.

sternr: btw @dwnusbaum , 2 questions about this code:
1. Why not use LambdaMetafactory instead of old fashioned reflection?
2. Can this entire block be run on a background thread (and managed with some sort of an ExecutorService?)

I'm trying to reduce the load this method takes

dwnusbaum: @sternr I am not familiar with the logic here, so I am not sure about questions 2 or 3, and I have not used `LambdaMetafactory` so I'm not sure about point 1. I think if you are seeing consistent performance issues with this method, it would be best to [open a Jira ticket](https://issues.jenkins-ci.org) that describes what you are doing and what issues you are seeing, and includes any thread dumps or other diagnostics that helped you identify an issue with this method. Feel free to ping me on any ticket you open (as `dnusbaum`) and I will try to get someone with a better understanding of this code to take a look.

PR #235, opened 2019-01-04
[JENKINS-55411] Whitelist DateTimeFormatter.ofPattern(String)

zhogov:
Whitelisted [DateTimeFormatter.ofPattern(String)](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ofPattern-java.lang.String-) since currently we have only set of constants whitelisted (BASIC_ISO_DATE, ISO_DATE_TIME...) that doesn't cover all real-life use cases

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,13 +8,13 @@
   </parent>
 
     <artifactId>script-security</artifactId>
-    <version>1.49</version>
+    <version>${revision}${changelist}</version>
     <packaging>hpi</packaging>
     <name>Script Security Plugin</name>
     <description>Allows Jenkins administrators to control what in-process scripts can be run by less-privileged users.</description>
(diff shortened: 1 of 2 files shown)


zhogov:
I have no idea why it shows 3 commits, I pushed only one

zhogov:
Oh, I see, because of force-push in master by **abayer**

dwnusbaum:
I think the problem is that your commit was based on commits in `master` that were force-pushed over for the recent security release. In other words, you built on top of the merge commit for https://github.com/jenkinsci/script-security-plugin/pull/234, but that commit no longer exists. If you rebase onto 6ff803006d (current `master`) then the extra commits should go away.

dwnusbaum:
Superseded by #237.

PR #240, opened 2019-03-22
[JENKINS-56688] Add missing Map utility methods.

daniel-carrington:
These methods are needed for general usage and scripting.
They are easily emulated with the already approved
methods containsKey and get.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -320,7 +320,6 @@ method java.util.Map putIfAbsent java.lang.Object java.lang.Object
 method java.util.Map remove java.lang.Object
 method java.util.Map replace java.lang.Object java.lang.Object
 method java.util.Map replace java.lang.Object java.lang.Object java.lang.Object
-method java.util.Map replaceAll java.lang.Object
 method java.util.Map size
 method java.util.Map values
 method java.util.Map$Entry getKey
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -311,11 +311,16 @@ method java.util.Map clear
 method java.util.Map containsKey java.lang.Object
 method java.util.Map entrySet
 method java.util.Map get java.lang.Object
+method java.util.Map getOrDefault java.lang.Object java.lang.Object
 method java.util.Map isEmpty
 method java.util.Map keySet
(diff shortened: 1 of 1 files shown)


daniel-carrington:
Relates to jenkinsci/workflow-cps-plugin#275 - using getOrDefault instead of Groovy 2-arg get is needed when working with unmodifiable maps such as the workflow `params` map.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -311,11 +311,16 @@ method java.util.Map clear
 method java.util.Map containsKey java.lang.Object
 method java.util.Map entrySet
 method java.util.Map get java.lang.Object
+method java.util.Map getOrDefault java.lang.Object java.lang.Object
```
jglick: FTR this is in lieu of the existing entry

```
staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods get java.util.Map java.lang.Object java.lang.Object
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -311,11 +311,16 @@ method java.util.Map clear
...
 method java.util.Map isEmpty
 method java.util.Map keySet
 method java.util.Map put java.lang.Object java.lang.Object
 method java.util.Map putAll java.util.Map
+method java.util.Map putIfAbsent java.lang.Object java.lang.Object
 method java.util.Map remove java.lang.Object
+method java.util.Map replace java.lang.Object java.lang.Object
+method java.util.Map replace java.lang.Object java.lang.Object java.lang.Object
+method java.util.Map replaceAll java.lang.Object
```
dwnusbaum: Looks like the test failure is because this method doesn't exist. There is `Map.replaceAll(BiFunction)`, but I'm not sure how well java's lambda functions interact with Groovy.

jglick: I think you meant `java.util.function.BiFunction`? See test failure.

daniel-carrington: Yes, that would've been better. I did not intend to add it, however - Java lambdas aren't nice single-function interfaces that play well with Groovy closures. I also avoided compute etc. which also require the lambda API.

daniel-carrington: I decided to remove it - java lambdas do not play well with Groovy.

daniel-carrington:
Looks good to merge now? @jglick @dwnusbaum

PR #241, opened 2019-04-05
whitelist java.lang.management.ManagementFactory.getOperatingSystemMXBean

grooverdan:
The read-only information returned provides useful information for architecture and OS, load information which could guide the necessary build/test.

All could be obtained with a bit more difficulty with shell scripts.

https://docs.oracle.com/javase/7/docs/api/java/lang/management/OperatingSystemMXBean.html


grooverdan:
Quite right, I'm after info on the agent rather than the master (and yes, different). I'm was intending to use this for architecture detection and maybe OS/Distro detection.

PR #242, opened 2019-04-05
Whitelist some methods related to java.time.Instant

grooverdan:
For https://stackoverflow.com/questions/49028088/scripts-not-permitted-to-use-staticmethod-java-time-instant-parse-java-lang-char users.


bitwiseman:
@grooverdan Could you comment on what ways this differs from `System.nanoTime`?

grooverdan:
Looking at the blacklist for nanoTime it references SECURITY-683 (which I can't find/access) with references to Spectre/Meltdown.

With that in mind only `staticMethod java.time.Instant now` is really a problem as the other are just accessors fields/maths operator implementations. The `java.time.Instant.now` method, like `java.time.LocalDate.now` (approved in #209) uses the system time, unlike `System.nanoTime` which uses much more precise CPU based implementation of timing. I'd assume its implementation is less precise which is why those where implemented.

Seems mozilla folks are happy if the timing resolution is >= 20us.
https://blog.mozilla.org/security/2018/01/03/mitigations-landing-new-class-timing-attack/
As best I can tell system time is in order of microseconds
https://bugs.openjdk.java.net/browse/JDK-8068730
So its order less than what Mozilla folks are happy with (probably quite conservative in the early days of Spectre) but its out of the nanoseconds relm for which System.nanoTime was blacklisted.

grooverdan:
`staticMethod java.time.Instant now` removed

dwnusbaum:
Rebuilding after #293 was merged.

PR #244, opened 2019-04-09
whitelist groovy.text.SimpleTemplateEngine

grooverdan:


grooverdan:
Thanks for the relevant links. I'll re-examine the use case.

PR #246, opened 2019-04-12
Always reject System.exit.

abayer:
There is never a valid reason for calling `System.exit(1)` in a
Jenkins Groovy script. Ever. So let's just permanently hack it into
always being rejected. Also, let's special case it in `ScriptApproval`
to not even show up as a pending signature.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java
@@ -52,6 +52,7 @@ import org.codehaus.groovy.control.Phases;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.Whitelist;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.ProxyWhitelist;
+import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.StaticWhitelist;
 import org.jenkinsci.plugins.scriptsecurity.scripts.ApprovalContext;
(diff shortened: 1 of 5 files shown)


daniel-beck:
What about `Runtime#exit` and `Runtime#halt`?

daniel-beck:
Honestly if something's blacklisted there's no reason to allow it in a sandbox. Want to shoot yourself in the foot, you need to be an admin for that.

abayer:
I've added `Runtime#exit(int)` and `Runtime#halt(int)`, and generalized the approach a bit to serve as the basis for a longer-term perma-blacklist solution.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -127,10 +127,10 @@
 
             // Allow calling closure variables from a script binding as methods
             if (receiver instanceof Script) {
-                Script s = (Script)receiver;
+                Script s = (Script) receiver;
```
jglick: Please avoid unrelated whitespace changes. Makes your diff hard to read.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java
```diff
@@ -121,7 +122,10 @@ public Scope enter() {
             if (ExtensionList.lookup(RootAction.class).get(ScriptApproval.class) == null) {
                 return; // running in unit test, ignore
             }
-            ScriptApproval.get().accessRejected(x, _context);
+            String signature = x.getSignature();
+            if (signature != null && !StaticWhitelist.isPermanentlyBlacklisted(signature)) {
```
jglick: Uh, what? No. You need to call `accessRejected` here even when `signature == null` (that would be an “unclassified” error).

PR #249, opened 2019-05-06
JENKINS-57344: Whitelist requests for some java.io and other misc calls

haridsv:
This PR whitelists a few java.io reader and writer methods that are useful on filters (wrapper readers and writers). It also whitelists a few list operations

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -40,25 +40,26 @@ new java.io.PrintWriter java.io.Writer
 method java.io.PrintWriter print java.lang.Object
 method java.io.PrintWriter print java.lang.String
 
+method java.io.Reader read
+method java.io.Reader read char[]
+method java.io.Reader read char[] int int
+method java.io.Reader reset
+method java.io.Reader skip long
+
 # Useful general stuff:
 new java.io.StringReader java.lang.String
-method java.io.StringReader close
-method java.io.StringReader read
-method java.io.StringReader read char[]
-method java.io.StringReader read char[] int int
-method java.io.StringReader reset
-method java.io.StringReader skip int
 new java.io.StringWriter
-method java.io.StringWriter append char
-method java.io.StringWriter append java.lang.CharSequence
-method java.io.StringWriter append java.lang.CharSequence int int
-method java.io.StringWriter close
-method java.io.StringWriter flush
-method java.io.StringWriter write char[]
-method java.io.StringWriter write char[] int int
-method java.io.StringWriter int
-method java.io.StringWriter write java.lang.String
-method java.io.StringWriter write java.lang.String int int
+
+method java.io.Writer write char[]
+method java.io.Writer write char[] int int
+method java.io.Writer write int
+method java.io.Writer write java.lang.String
+method java.io.Writer write java.lang.String int int
+
+method java.lang.Appendable append char
+method java.lang.Appendable append java.lang.CharSequence
+method java.lang.Appendable append java.lang.CharSequence int int
+method java.lang.AutoCloseable close
 new java.lang.Boolean java.lang.String
 staticMethod java.lang.Boolean parseBoolean java.lang.String
 staticMethod java.lang.Boolean valueOf boolean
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -34,13 +34,31 @@ method groovy.lang.Closure setResolveStrategy int
...
 
+# Useful to show stacktrace to the user.
+new java.io.PrintWriter java.io.Writer
 # Needed for Groovy Templates to emit output:
 method java.io.PrintWriter print java.lang.Object
 method java.io.PrintWriter print java.lang.String
 
 # Useful general stuff:
 new java.io.StringReader java.lang.String
+method java.io.StringReader close
```
jglick:
```
Signatures out of order, at index 34 found new java.io.StringReader java.lang.String but expected method java.io.StringReader close
Signatures out of order, at index 35 found method java.io.StringReader close but expected new java.io.StringReader java.lang.String
Signatures out of order, at index 47 found method java.io.StringWriter write char[] but expected method java.io.StringWriter int
Signatures out of order, at index 48 found method java.io.StringWriter write char[] int int but expected method java.io.StringWriter write char[]
Signatures out of order, at index 49 found method java.io.StringWriter int but expected method java.io.StringWriter write char[] int int
Signatures out of order, at index 293 found new java.util.LinkedHashSet but expected method java.util.Iterator hasNext
Signatures out of order, at index 294 found method java.util.Iterator hasNext but expected method java.util.Iterator next
Signatures out of order, at index 295 found method java.util.Iterator next but expected new java.util.LinkedHashSet
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -34,13 +34,31 @@ method groovy.lang.Closure setResolveStrategy int
...
+method java.io.StringReader read
```
jglick: You rather want to whitelist the _defining_ method, in `Reader`. Ditto in other cases.

bitwiseman:
@haridsv Do you plan to update this PR?

haridsv:
@bitwiseman Apologies for the delay, I will try to look into this tomorrow. I need to run this test locally to see the error and I think I did that a long time ago just need to find the steps again.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
haridsv: Missed to respond to this earlier. That would open up those methods for all readers, right? Won't that whitelist too much, e.g., reading local files etc.?

haridsv: Looks like specific overridden methods can't be whitelisted, so it is all or nothing.

haridsv: I pushed a new change with the right attribution. I am assuming it is still OK because as long as the sandbox prevents the construction of those objects, there is not going to be a security issue with the broader whitelisting.

jglick: > reading local files

You would first need to construct a `FileReader`, which would not be allowed.

haridsv:
Some whitelist behavior seems to have changed. Previously, something like `"/a/b/c".split("/")[0..-2]` failed because of lack of whitelist for `getAt java.lang.Object[] groovy.lang.Range`, so I added that in PR. Now, the exact same Groovy snippet complains about the lack of whitelist for a different signature, which is `getAt java.lang.Object[] java.util.Collection`. It looks like the parameter type is `groovy.lang.IntRange` and it implements both `groovy.lang.Range<Integer>` and `java.util.Collection<Integer>`, so how is the interface chosen for creating the signature? Why did it switch from using `Range` to `Collection` ? Is this an indication of a bug in creating the signature in a deterministic way?

haridsv:
@dwnusbaum How to mark the jira as fixed? I probably don't have privileges to edit that field.

PR #250, opened 2019-05-09
[JENKINS-57563] - Added JCasC to ScriptApproval (PR#233)

pjanouse:
This is a follow-up of #233


src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/JcascTest.java
```diff
@@ -0,0 +1,31 @@
...
+import static org.junit.Assert.assertTrue;
+
+public class JcascTest {
+
+    @Rule
+    public JenkinsConfiguredWithCodeRule j = new JenkinsConfiguredWithCodeRule();
+
+    @Test
+    @ConfiguredWithCode("smoke-test-empty.yaml")
+    public void smokeTestEmpty() throws Exception {
```
timja: seems unneeded to me? not really relevant to testing jcasc config

src/test/resources/org/jenkinsci/plugins/scriptsecurity/scripts/smoke-test-entry.yaml
```diff
@@ -0,0 +1,8 @@
+unclassified:
+  scriptApproval:
+    approvedSignatures:
+    - method java.net.URI getHost
+
+jenkins:
```
timja: neither of these are relevant to the test I recommend deleting

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/JcascTest.java
```diff
@@ -0,0 +1,31 @@
...
+import io.jenkins.plugins.casc.misc.ConfiguredWithCode;
+import io.jenkins.plugins.casc.misc.JenkinsConfiguredWithCodeRule;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class JcascTest {
```
timja: Could you add an export test too please?
https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#model-export-test

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/JcascTest.java
```diff
@@ -0,0 +1,38 @@
...
+        String[] approved = ScriptApproval.get().getApprovedSignatures();
+        assertTrue(approved.length == 1);
+        assertEquals(approved[0], "method java.net.URI getHost");
+    }
+
+    @Test
+    public void smokeTestExport() throws Exception {
+        ConfiguratorRegistry registry = ConfiguratorRegistry.get();
+        ConfigurationContext context = new ConfigurationContext(registry);
+        CNode yourAttribute = Util.getUnclassifiedRoot(context).get("scriptApproval");
```
timja: I would suggest adding a static import for `Util`
Its intended to be done that way so that this reads better

```suggestion
        CNode yourAttribute = getUnclassifiedRoot(context).get("scriptApproval");
```

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/JcascTest.java
```diff
@@ -0,0 +1,38 @@
...
+        String exported = Util.toYamlString(yourAttribute);
```
timja:
```suggestion
        String exported = toYamlString(yourAttribute);
```

pjanouse:
Added export test and refactored the rest, I think at least smoke test with explicitly defined approved signature entry should remain...

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/JcascTest.java
```diff
@@ -0,0 +1,38 @@
...
+        String expected = Util.toStringFromYamlFile(this, "smoke_test_expected.yaml");
```
timja:
```suggestion
        String expected = toStringFromYamlFile(this, "smoke_test_expected.yaml");
```

src/test/resources/org/jenkinsci/plugins/scriptsecurity/scripts/smoke_test.yaml
```diff
@@ -0,0 +1,8 @@
+unclassified:
+  scriptApproval:
+    approvedSignatures:
+    - method java.net.URI getHost
+
+jenkins:
```
timja: I recommend removing this and the next 2 lines as its not relevant

timja:
@dwnusbaum / @abayer / @jglick 
Adding you as you were on the previous PR: https://github.com/jenkinsci/script-security-plugin/pull/233

mind reviewing please?

casz:
We need JCasC 1.15 for tests to pass 😓 I'll get a release going

pjanouse:
Looks no luck with my attempt to resolve the issue we've seen on MS Windows platform - what it is exactly? Platform specific EOL issue?

timja:
If you update the configuration-as-code version it should be sorted, we now replace `\r\n` with just `\n`

The other way to work around it is to add a .gitattributes entry for yaml files

pjanouse:
Guys, I'm happy with current state.

casz:
@dwnusbaum @abayer @jglick do you mind reviewing?

oleg-nenashev:
Created https://issues.jenkins-ci.org/browse/JENKINS-57563 for this ticket

rafaljanicki:
Hey Guys, can we have it merged? Thanks

oleg-nenashev:
It basically needs feedback from @dwnusbaum @abayer @kshultzCB @bitwiseman and other @jenkinsci/script-security-plugin-developers . We cannot proceed without it

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -68,16 +71,18 @@
...
+import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.bind.JavaScriptMethod;
 
 /**
  * Manages approved scripts.
  */
-@Extension public class ScriptApproval implements RootAction, Saveable {
+@Symbol("scriptApproval")
+@Extension
+public class ScriptApproval extends GlobalConfiguration implements RootAction {
```
oleg-nenashev: Should it be added to the `Security` category so that it is not `unclassified` ?

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -668,6 +655,14 @@ public static void popRegistrationCallback() {
         callbacks.get().pop();
     }
 
+    @DataBoundSetter
+    public synchronized void setApprovedSignatures(String[] signatures) {
```
oleg-nenashev: What about classpaths & Co?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/config.jelly
```diff
@@ -0,0 +1,3 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:l="/lib/layout" xmlns:f="/lib/form">
```
oleg-nenashev: Does it appear in Web UI with such empty config? IIUC it may be confusing to user and, what is worse, cause undesired configuration overrides when submitting config pages

timja: In @dwnusbaum's review he tested it and said it worked fine:
https://github.com/jenkinsci/script-security-plugin/pull/233#pullrequestreview-179903699

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
pjanouse: I think this can be tracked as further improvement separately. Current semantic can be extended in the future to cover this as well. WDYT?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/config.jelly
pjanouse: So can we close this as resolved?

casz:
@pjanouse Andrew's points are valid, do you have time to move forward with this?

pjanouse:
I'm not sure I understand all items which are requested (I'm not the author of the change request, I've only created the new PR from #233 and have been trying to moved this forward), but I'll do my best definitely.

olivergondza:
@abayer, can you be more specific on you would like to see tested as a part of this PR? To me, it is expected that permitted signatures got replaced by what is declared by JCasC and that those signatures takes effect for running jobs (former is tested, letter is not (do we need it?)). But I presume that you are after some sort of merging of existing and declared approvals which does not feel right to me to begin with...

pom.xml
```diff
@@ -57,6 +58,19 @@
           <artifactId>groovy</artifactId>
         </exclusion>
       </exclusions>
+    </dependency>
+        <dependency>
```
casz: odd indent

pjanouse: Ups...

placydo:
Hey any updates on it?

dwnusbaum:
> can you be more specific on you would like to see tested as a part of this PR? To me, it is expected that permitted signatures got replaced by what is declared by JCasC and that those signatures takes effect for running jobs (former is tested, letter is not (do we need it?)).

My interpretation of @abayer's ask (which I agree with) was for a test using [`@LocalData`](https://github.com/jenkinsci/jenkins-test-harness/blob/3f11e2d537a71c6d423829b3764157d561043632/src/main/java/org/jvnet/hudson/test/recipes/LocalData.java) with signatures configured based on the most recently released version of the plugin to make sure that after upgrading to a version of the plugin with the changes in this PR `scriptApproval.xml` still loads correctly (i.e., that this PR doesn't break users who aren't using CasC). Something like this:

```
@Issue("JENKINS-57563")
@LocalData // Data is just a non-empty `scriptApproval.xml` with some custom entries that are not in the standard whitelist
@Test
public void upgradeSmokes() throws Exception {
    FreeStyleProject p = r.createFreeStyleProject();
            new SecureGroovyScript("some script that requires the approvals in scriptApproval.xml", true, null)));
    FreeStyleBuild b = r.buildAndAssertSuccess(p);
    // other assertions as necessary/useful to show the script ran successfully
}
```

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest.java
```diff
@@ -136,6 +143,22 @@ public void malformedScriptApproval() throws Exception {
         assertEquals(0, sa.getDangerousApprovedSignatures().length);
     }
 
+    @Issue("JENKINS-57563")
+    @LocalData("upgrade")
```
dwnusbaum: FWIW if you rename the zip file to `upgradeSmokes` to match the name of the test then you don't need to specify a name here. I might also add a small note like: `// Just a scriptApproval.xml that whitelists 'staticMethod jenkins.model.Jenkins getInstance'`

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest.java
```diff
@@ -136,6 +143,22 @@ public void malformedScriptApproval() throws Exception {
...
+    @Test
+    public void upgradeSmokes() throws Exception {
+        r.jenkins.setSecurityRealm(r.createDummySecurityRealm());
+        r.jenkins.setAuthorizationStrategy(new MockAuthorizationStrategy().
```
dwnusbaum: I don't think Jenkins' security realm or authorization strategy matter for this test since we are running in the sandbox, so these lines could just be deleted.

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest.java
casz: You should be able to use the xml by it self instead of using a zip file
See: https://github.com/jenkinsci/matrix-auth-plugin/pull/48/files for comparison

dwnusbaum: Yeah good point, for a simple case like this just having the file specified directly might be easier.

pjanouse: I've tried what @dwnusbaum suggested, it works as expected, so I'm including it in my last set of commits. 
I've also tried to deal with the approach advised by @casz, but I had to do sth wrong. I placed the file with the same content as in the ZIP into resources/.../ScriptApprovalTest/security/config.xml, but it doesn't work. I've tried several other changes, but nothing works for me. If we really need to remove the 'ZIP approach', could you advise me where exactly and which content should be placed? In that case we should change also other tests in ScriptApprovalTest to unify that, thank you.

dwnusbaum: @pjanouse The zip approach is fine, it has the same behavior as what @casz is recommending, the benefit is just that it would be easier for anyone browsing the source code to understand the test without needing to unzip the archive. I think the resource path you would want would be `src/main/resources/.../ScriptApprovalTest/upgradeSmokes/scriptApproval.xml`, i.e. you just unzip your archive into a directory with the same name. If that isn't working for you, no worries, we can leave it as is.

pjanouse: Hmm, I think this was one of my un-successful attempt... tried again and it works as expected, so using...

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/config.jelly
dwnusbaum: Things work fine with the file, but as far as I can tell everything works fine without it. Is there a reason it was added? If not, it should be deleted.

dwnusbaum: I guess it is just there to satisfy `GlobalConfiguration.getGlobalConfigPage`. In that case maybe better to leave it, although could be good to add a comment here explaining why the empty file exists?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/config.jelly
```diff
@@ -0,0 +1,4 @@
+<!-- Just ti satisfy GlobalConfiguration.getGlobalConfigPage() -->
```
casz:
```suggestion
<!-- Just to satisfy GlobalConfiguration.getGlobalConfigPage() -->
```

chrismaes87:
Would be really nice to get this merged. This is the only issue blocking me from setting up end-to-end test for our continuous integration pipeline.

I have tried quite some workarounds without success: https://stackoverflow.com/questions/56938563/how-to-avoid-jenkins-asking-for-in-process-approval-for-call-in-trusted-shared-p

keep up the good work!

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -668,6 +661,14 @@ public static void popRegistrationCallback() {
...
+        Jenkins.getInstance().checkPermission(Jenkins.RUN_SCRIPTS);
+        approvedSignatures.clear();
+        approvedSignatures.addAll(Arrays.asList(signatures));
+        save();
```
dwnusbaum: I just realized that we never actually reconfigure `ApprovedWhitelist` based on these signatures until a sandboxed script runs for the first time. I think you should add a call to `reconfigure` here like the other methods that alter `approvedSignatures`.

Also, unlike the other methods that take signatures like `approveSignature`, in `setApprovedSignatures` users have direct control over the signature itself, so it might also be a good idea to do something like this to filter out any totally malformed signatures:

```
List<String> goodSignatures = new ArrayList<>(signatures.length);
for (String signature: signatures) {
  try {
    StaticWhitelist.parse(signature); // Method visibility might need to be changed to package-private
    goodSignatures.add(signature);
  } catch (IOException e) {
    LOGGER.log(Level.WARNING, "Ignoring malformed signature: " + signature);
  }
  ... the rest of the method ...
}
```

If you don't do this (but you do add the call to `reconfigure` after save), and a malformed signature does end up being saved, then approval of signatures from the UI will be broken until the malformed signature is removed because an exception will be thrown when `reconfigure` is called from `approveSignature` or similar.

pjanouse: Good idea, I'll fix that...

Rahaman148:
@casz 
hi,
 can I use this yaml code for script approval without any user interaction. Right now whenever I changed or modified groovy scripts it always ask for `in process script approval`, during automated build not possible to do everytime go to Jenkins UI and click on `Administrator...script approval...`. here is the code which I found #233 
```
yaml
---
unclassified:
  scriptApproval:
    approvedSignatures:
    - method java.net.URI getHost
    - method java.net.URI getPort
    - new java.net.URI java.lang.String

```
Any help please

Rahaman148:
@timja @ewelinawilkosz
can you please update does yaml support script approval, If so could you please share that code. I am stuck at this point. I found some groovy script which didn't help. I mentioned details on top.

Any help will highly appreciated

casz:
This PR tries to add the support @Rahaman148
Without this being merged there is no support.

@oleg-nenashev @abayer @bitwiseman @kshultzCB would get to get this unstuck.
The PR satisfy the last review by @abayer

dwnusbaum:
Sorry, this fell off my radar when the recent [groovy-sandbox/script-security vulnerabilities](https://jenkins.io/security/advisory/2019-07-31/#SECURITY-1465%20(1)) came up. I'll try to get this merged in the next week or so.

Rahaman148:
@dwnusbaum 
can you please tell me when it will be merged on that time it will support this feature automatically or not , and sometimes it display messages something like this during build job in jenkins UI `Administer...can whether decide to approve this method or signature or not`

scriptApproval

I want to know this yaml will support on above feature or not or if you've any updated yaml code please share it thanks.
```
yaml
---
unclassified:
  scriptApproval:
    approvedSignatures:
    - method java.net.URI getHost
    - method java.net.URI getPort
    - new java.net.URI java.lang.String

```
will highly appreciate your help and response

casz:
@dwnusbaum would really appreciate if this could late with jcasc v1.30

pom.xml
```diff
@@ -7,29 +7,30 @@
...
+  <packaging>hpi</packaging>
+  <name>Script Security Plugin</name>
+  <description>Allows Jenkins administrators to control what in-process scripts can be run by less-privileged users.</description>
+  <url>https://wiki.jenkins.io/display/JENKINS/Script+Security+Plugin</url>
+  <properties>
+    <revision>1.62</revision>
+    <changelist>-SNAPSHOT</changelist>
+    <jenkins.version>2.60.3</jenkins.version>
+    <java.level>8</java.level>
+    <configuration-as-code.version>1.15</configuration-as-code.version>
```
timja:
```suggestion
    <configuration-as-code.version>1.30</configuration-as-code.version>
```

pjanouse: +1, it was the latest at the time of creation

pjanouse:
Rebased against current master as well

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelist.java
```diff
@@ -156,7 +156,7 @@ public static boolean isPermanentlyBlacklistedConstructor(@Nonnull Constructor c
      * @return the equivalent {@link Signature}
      * @throws IOException if the signature string could not be parsed.
```
oleg-nenashev:
```suggestion
     * @throws IOException if the signature string could not be parsed.
     * @since TODO
```

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -668,6 +661,24 @@ public static void popRegistrationCallback() {
...
+    public synchronized void setApprovedSignatures(String[] signatures) throws IOException {
+        Jenkins.getInstance().checkPermission(Jenkins.RUN_SCRIPTS);
+        approvedSignatures.clear();
+        List<String> goodSignatures = new ArrayList<>(signatures.length);
+        for (String signature : signatures) {
+            try {
+                StaticWhitelist.parse(signature);
+                goodSignatures.add(signature);
+            } catch (IOException e) {
+                LOG.warning("Ignoring malformed signature: " + signature);
```
oleg-nenashev: Add exception to the warning so that there is some diagnostics?

rsandell:
@pjanouse Please stop force pushing, it makes re-reviewing very very hard!

pjanouse:
Sure, I don'l like it too, but if I rebase this PR against current master, I can't do a simple merge/push. Is there another way how to achieve the same without force push (I don't know it yet, sorry)?

timja:
> Sure, I don'l like it too, but if I rebase this PR against current master, I can't do a simple merge/push. Is there another way how to achieve the same without force push (I don't know it yet, sorry)?

```
git merge master
git push
```

casz:
> ```
> git merge master
> git push
> ```

In hopes this repo uses squash 😆

pjanouse:
Ok, if 'merge master' is a preferred way instead of put new changes on top of it ('rebase'), I'll do it next time.

Some other projects (outside of Jenkins) have another 'workflow' how to handle that.

rsandell:
@pjanouse @dwnusbaum I fixed the merge conflict in pom.xml to try to expedite the review.

rsandell:
While I was at it I also fixed the whitespace changes to match master, even though they are technically wrong compared to the rest of the pom file the diff gets smaller that way and hopefully won't give merge conflicts to other PRs.

timja:
Is this ready to go? 🚢  😄

casz:
Would be nice to :shipit:

batmat:
Thanks a lot for helping move this forward @rsandell !

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/config.jelly
dwnusbaum: nit: Looks like the typo is still there.

dwnusbaum:
Ok, merged and released in script-security 1.64. Thanks @dgarzon @pjanouse and @rsandell for working on it!

cwiggs:
I tried to use `scriptApproval` in Jenkins today and I think I found a bug that I reported in JENKINS-59979

It seems like the approvedSignatures aren't actually applied until Jenkins is restarted, was anyone else able to get this working without restarting Jenkins?

PR #160, opened 2017-10-31
[JENKINS-47430] Rip out Guava Cache and replace with Caffeine

svanoort:
Attempt to fix some bizarre test hangs tied to Guava's concurrent loading implementation, and also to improve performance (we've noted cache-internal operations coming up fairly often in stack traces for high-scale pipeline instances). 

[Sample Error](https://gist.githubusercontent.com/abayer/068f1b96a6368a75e17a7871914d380f/raw/babc61336878b11728fc69b0dc8cf823df8febc3/gistfile1.txt). 

[Performance-side, a JIRA link](https://issues.jenkins-ci.org/browse/JENKINS-47430). 

Caffeine uses its own strategy, which is [much higher performance](https://github.com/ben-manes/caffeine/wiki/Benchmarks). 

Now, the catch for this: we need to update to Java 8 -- but since the 2.60.x LTS line is 3 months old, that's not the end of the world.

Edit: deployed snapshot as 1.35-20171031.175306-1

Edit: if anything, seems to be slightly slower, but probably that can be solved by restructuring some of the methods here since they're not as efficiently JIT-able as they might be.

PR #252, opened 2019-05-23
SandboxResolvingClassLoader.parentClassCache could leak loaders

jglick:
Tracked down a memory leak in `SandboxResolvingClassLoader.parentClassCache`. Turns out its use of `weakKeys` did not work at all. The existing test did not catch this, since the `parentLoader` in that case is an `UberClassLoader`, which you would not expect to be collected anyway. But if you have a nonempty `SecureGroovyScript.classpath`, for example, then the parent loader was a new `URLClassLoader`, and this never got released. Thus any classes actually loaded from a classpath JAR would never be released, leading to increased heap/metaspace on each build.

(Not sure if there is a case where this could affect Pipeline; the observed case involved the `uno-choice` plugin and build parameters on a freestyle project. https://github.com/jenkinsci/jenkins/pull/4035 also came out of this case; [CloudBees-internal issue](https://cloudbees.atlassian.net/browse/CPLT2-5589).)

Subsumes #251. Merge conflict with #160, which would be a somewhat bigger change.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
@@ -1,12 +1,16 @@
 package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
 
 import com.google.common.base.Optional;
+import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 import groovy.lang.GroovyShell;
 import java.net.URL;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
```diff
@@ -65,33 +62,37 @@
...
-        return CacheBuilder.newBuilder().weakKeys().build(new CacheLoader<ClassLoader,LoadingCache<String, T>>() {
-            @Override public LoadingCache<String, T> load(final ClassLoader parentLoader) {
-                return CacheBuilder.newBuilder()./* allow new plugins to be used, and clean up memory */expireAfterWrite(15, TimeUnit.MINUTES).build(new CacheLoader<String, T>() {
-                    @Override public T load(String name) {
-                        Thread t = Thread.currentThread();
-                        String origName = t.getName();
-                        t.setName(origName + " loading " + name);
-                        long start = System.nanoTime(); // http://stackoverflow.com/q/19052316/12916
-                        try {
-                            return function.compute(parentLoader, name);
```
dwnusbaum: So if I understand correctly, the problem was that here we close over `parentLoader` in the inner `LoadingCache`, so `weakKeys` didn't do anything because the inner `LoadingCache` always had a strong reference to the parent loader? And in the new code, we instead pass a reference to the parent loader via a `Supplier` that is used to load the value if needed and is not retained anywhere?

jglick: > the problem was …

Exactly. The use of inner classes vs. `static` (nested) classes is not significant—the inner `LoadingCache` really needed a reference to the `parentLoader` in order to function.

> in the new code, we instead pass a reference to the parent loader

Well `load` here is just a helper method to avoid code duplication between the two caches, so the `Supplier` is not interesting. The key difference is that we are now using a plain `Cache` for the inner cache, which does not need to hold on to a `CacheLoader`, so there is no reference chain from the singleton outer cache to a `ClassLoader`. Instead, the `ClassLoader` is held only via local variable from a `loadClass` or `getResource` method, and `Cache.get(K,                       Callable<? extends V>)` is used instead of `LoadingCache.get(K)`.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyMemoryLeakTest.java
```diff
@@ -35,18 +39,19 @@ public void clearLoaders() {
...
 
     @Test
     public void loaderReleased() throws Exception {
-        Assume.assumeTrue(isRunningOnJDK8());
-
         FreeStyleProject p = r.jenkins.createProject(FreeStyleProject.class, "p");
+        String cp = CharacterPredicates.class.getProtectionDomain().getCodeSource().getLocation().toString(); // some JAR
         p.getPublishersList().add(new TestGroovyRecorder(
-                new SecureGroovyScript(GroovyMemoryLeakTest.class.getName()+".register(this)", false, null)));
+            new SecureGroovyScript(GroovyMemoryLeakTest.class.getName()+".register(this)", false, Collections.singletonList(new ClasspathEntry(cp)))));
```
bitwiseman: This assert should be verifying that the loader was released.  Okay.  

Should we still have a negative test for the the `PluginManager.UberClassLoader` where the release shouldn't/doesn't happen?

jglick: It is not possible for the `Jenkins.pluginManager.uberClassLoader` to be released.

bitwiseman:
I'll publish today.

PR #253, opened 2019-05-30
Classpath leak redux

jglick:
#252 was not enough; there was still a leak through `Class` values in the cache. Also part of https://github.com/jenkinsci/workflow-cps-plugin/pull/289 needed to be ported (and adapted) in order to clear some soft references in JAR classpath mode.


src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyMemoryLeakTest.java
```diff
@@ -49,9 +48,10 @@ public static void register(Object o) {
     @Test
     public void loaderReleased() throws Exception {
         FreeStyleProject p = r.jenkins.createProject(FreeStyleProject.class, "p");
-        String cp = CharacterPredicates.class.getProtectionDomain().getCodeSource().getLocation().toString(); // some JAR
-        p.getPublishersList().add(new TestGroovyRecorder(
-            new SecureGroovyScript(GroovyMemoryLeakTest.class.getName()+".register(this)", false, Collections.singletonList(new ClasspathEntry(cp)))));
+        String cp = GroovyMemoryLeakTest.class.getResource("somejar.jar").toString();
```
dwnusbaum: So the problem with the test was that the previous class wasn't actually loaded from a separate jar?

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -151,36 +151,49 @@ public SecureGroovyScript configuringWithNonKeyItem() {
...
+            for (Class<?> clazz : gcl.getLoadedClasses()) {
+                cleanUpClass(clazz, encounteredLoaders, encounteredClasses);
+            }
+            gcl.clearCache();
+        } else if (loader instanceof SandboxResolvingClassLoader) {
+            // OK, just check its parent
+        } else if (loader instanceof ClasspathURLClassLoader) {
+            Collection<Class<?>> loadedClasses = ((ClasspathURLClassLoader) loader).loadedClasses;
+            synchronized (loadedClasses) {
+                loadedClasses = new ArrayList<>(loadedClasses);
```
dwnusbaum: Do you need to call `ClasspathURLClassLoader.loadedClasses.clear()` here? It's not obvious to me why the collection of strong references to the classes in `ClasspathURLClassLoader` is not problematic. Is it because we are only concerned about external references in caches, etc., and so any references from the class loaders to the classes they loaded are fine so long as the class loader is not strongly referenced anywhere?

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
```diff
@@ -39,14 +40,19 @@
...
                 } catch (ClassNotFoundException x) {
                     return Optional.absent();
                 }
-            }).orNull();
-            if (c != null) {
+            });
+            if (ref.isPresent()) {
+                Class<?> c = ref.get().get();
+                if (c == null) {
+                    // Probably impossible for the class to be collected if its loader was not, but just in case…
```
dwnusbaum: Maybe worth a FINE-level log message? Would be nice to have an easy way to specify behavior like "log at FINE initially, but if you see this more than X times over an X-TimeUnit period, log at WARNING" in case something goes wrong and this is called way more often than expected.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
```diff
@@ -26,7 +27,7 @@
     
     private static final Logger LOGGER = Logger.getLogger(SandboxResolvingClassLoader.class.getName());
 
-    private static final LoadingCache<ClassLoader, Cache<String, Optional<Class<?>>>> parentClassCache = makeParentCache();
+    private static final LoadingCache<ClassLoader, Cache<String, Optional<WeakReference<Class<?>>>>> parentClassCache = makeParentCache();
```
dwnusbaum: Why not just use `weakValues` on the inner cache? Is it so we can tell the difference between a value that was not in the cache at all and a value that has been GC'd while in the cache so that we can fall back to the non-cached load? Wouldn't `weakValues` give us that behavior for free since the value would be loaded and put in the cache in both cases?

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyMemoryLeakTest.java
jglick: There were two problems with the previous revision of the test:

* The test did not actually load any classes from the specified classpath. So, it reproduced a leak when creating the outer cache, but not when adding entries to the inner cache. (Or perhaps dealt only with the resource cache, which does not suffer from the values-refers-to-key problem specific to the class cache fixed here, since a `URL` holds no interesting references.)
* The JAR file I used (`commons-text` or something) was in the test classpath (Maven `test` scope), so any attempts to load classes from it would actually go through `sun.misc.AppClassLoader` instead.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
jglick: `ClassLoader` already has

```java
// The classes loaded by this class loader. The only purpose of this table
// is to keep the classes from being GC'ed until the loader is GC'ed.
private final Vector<Class<?>> classes = new Vector<>();
```

but it is inaccessible, so we need a custom subclass that makes a similar list accessible. `GroovyClassLoader` happens to have such a method already, which we were using. Also having a custom subclass helps us identify a loader created earlier by `SecureGroovyScript` (with a nonempty classpath), so that we know when to terminate the recursive `parent` search (normally it ends at `UberClassLoader` but the API does permit a plugin to pass in some other base loader, typically a `PluginClassLoader`).

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
jglick: I guess it could be an assertion, but it did not seem worth the bother.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
jglick: Yes `weakValues` might simplify the patch, I will check it.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
```diff
@@ -89,8 +97,12 @@
         }
     }
 
-    private static <T> LoadingCache<ClassLoader, Cache<String, T>> makeParentCache() {
-        return CacheBuilder.newBuilder().weakKeys().build(new CacheLoader<ClassLoader, Cache<String, T>>() {
+    private static <T> LoadingCache<ClassLoader, Cache<String, T>> makeParentCache(boolean weakValues) {
+        CacheBuilder<Object, Object> builder = CacheBuilder.newBuilder().weakKeys();
```
jglick: Will again be a merge conflict with #160.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
```diff
@@ -26,27 +26,35 @@
...
 
     SandboxResolvingClassLoader(ClassLoader parent) {
         super(parent);
     }
 
+    /**
+     * Marker value for a {@link ClassNotFoundException} negative cache hit.
+     * Cannot use null, since the cache API does not permit null values.
+     * Cannot use {@code Optional<Class<?>>} since weak values would mean this is always collected.
+     * This value is non-null, guaranteed not to be a legitimate return value, and strongly held.
```
dwnusbaum: Why is `Void.class` guaranteed to not be a legitimate return value? What if someone uses `Void.class` explicitly in a Groovy script?

jglick: Ah, true, I was thinking of defining loaders but indeed this cache also applies to the results of all loads including parent delegation. Just a moment…

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
bitwiseman: Easy fix.

jglick: Yes (assuming Caffeine also support `weakValues`), just FYI so it does not come as a surprise.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
```diff
@@ -26,27 +26,37 @@
...
 
     SandboxResolvingClassLoader(ClassLoader parent) {
         super(parent);
     }
 
+    /**
+     * Marker value for a {@link ClassNotFoundException} negative cache hit.
+     * Cannot use null, since the cache API does not permit null values.
+     * Cannot use {@code Optional<Class<?>>} since weak values would mean this is always collected.
+     * This value is non-null, not to be a legitimate return value
```
dwnusbaum:
```suggestion
     * This value is non-null, not a legitimate return value
```

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
bitwiseman: It does.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyMemoryLeakTest.java
jglick: > reproduced a leak when creating the outer cache, but not when adding entries to the inner cache. (Or perhaps dealt only with the resource cache…

Amendment: the class cache normally has a bunch of entries, including positives like `Object` or `WorkflowScript` as well as negatives for all sorts of Groovy nonsense. But in the previous test, none of the positive entries were for classes actually defined in the `URLClassLoader`—they were defined by lower loaders—so the fact that they were strongly held from the cache was irrelevant.

jglick: To make it clearer still: when you actually load a class _from_ the classpath, there was until this fix a routine mistake of the sort that the `WeakHashMap` Javadoc warns about: a weakly held `URLClassLoader` key but then a strongly held `Class` value which in turn strongly holds the key. Pending my longstanding favorite [JDK-6389107](https://bugs.openjdk.java.net/browse/JDK-6389107), there is no easy universal fix for this problem. Nor is there any feasible way that I am aware of to mechanically detect the mistake at runtime (some profilers detect it for `WeakHashMap`, probably not for Guava caches); the `MemoryAssert` shown here can find it, if you think to write the right test!

PR #254, opened 2019-05-30
PoC of JDK-6389107 workaround

jglick:
Demonstrating that the special case of a `ClassLoader`- (or `Class-`) keyed cache can be handled despite [JDK-6389107](https://bugs.openjdk.java.net/browse/JDK-6389107) if you are willing to poke around a tiny bit in Java internals.

I am not inclined to merge this as `SandboxResolvingClassLoader.parentClassCache` from #252 / #253 can do without (weakly holding values is OK for that use case), but wanted to record this in case we come across another situation where we are otherwise stuck with `SoftReference`d values, manual cache cleanup (e.g. using `AutoCloseable`), or other undesirable workarounds. Certainly Groovy itself and `java.beans.Introspector` could use a utility like this in numerous places (and save us from the `cleanUpHeap` hacks).

No unit tests at this point, just tested implicitly via `GroovyMemoryLeakTest` etc.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/ClassLoaderCache.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/ClassLoaderCache.java
@@ -1,0 +1,102 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2019 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
```diff
@@ -72,42 +57,29 @@
...
-                    return supplier.get();
-                } finally {
-                    t.setName(origName);
-                    long ms = (System.nanoTime() - start) / 1000000;
-                    if (ms > 1000) {
-                        LOGGER.log(Level.INFO, "took {0}ms to load/not load {1} from {2}", new Object[] {ms, name, parentLoader});
+    private static <T> ClassLoaderCache<LoadingCache<String, T>> makeParentCache(BiFunction<ClassLoader, String, T> func) {
+        // TODO expire cache if new plugins are installed
+        return new ClassLoaderCache<>(parentLoader -> {
+            return CacheBuilder.newBuilder().build(new CacheLoader<String, T>() {
```
jglick: Not using anything interesting from Guava here; could just replace with, say, `ConcurrentHashMap.computeIfAbsent`.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/ClassLoaderCache.java
```diff
@@ -0,0 +1,102 @@
...
+ * @see <a href="https://bugs.openjdk.java.net/browse/JDK-6389107">JDK-6389107</a>
+ */
+@SuppressWarnings("unchecked")
+final class ClassLoaderCache<V> {
+
+    private static final Object HACK_KEY = ClassLoaderCache.class.getName() + "$$__$$"; // not a valid package name
+
+    /**
+     * Need a {@link Map}-valued instance field in {@link ClassLoader} which is never null and whose entries are never enumerated.
+     * We will stash a mistyped value in there under a special {@link #HACK_KEY}.
```
dwnusbaum: Looks like something from the Jigsaw team's nightmares 😄

jglick:
Anyway. Served its function for now. Something to put in our back pocket.

PR #258, opened 2019-06-04
Whitelist Number.times(Closure)

jglick:
Handy for scripts like

```groovy
10.times {
  // …
}
```

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -672,6 +672,7 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods take java.lang.Ite
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods take java.util.List int
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods takeRight java.lang.Iterable int
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods takeRight java.util.List int
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods times java.lang.Number groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toBoolean java.lang.String
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toInteger java.lang.String
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toList boolean[]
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -401,17 +401,17 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods collectNested java
...
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods contains char[] java.lang.Object
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods contains double[] java.lang.Object
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods contains float[] java.lang.Object
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods contains int[] java.lang.Object
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods contains java.lang.Iterable java.lang.Object
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods contains java.lang.Object[] java.lang.Object
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods contains long[] java.lang.Object
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods contains short[] java.lang.Object
-staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods containsAll Collection[] java.lang.Object
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods containsAll java.lang.Iterable java.lang.Object[]
```
jglick: Mistakes from #137 that I just noticed while I was here. Currently `StaticWhitelistTest.sanity` tolerates `ClassNotFoundException` since there are some entries corresponding to classes not accessible from the test classpath of this plugin.

jglick: (though maybe these ought to be enumerated)

dwnusbaum: Yes, I was thinking that maybe we could split the whitelist in two and strengthen the checks on the whitelist where everything is available, but it doesn't seem critical.

jglick: > split the whitelist in two

Actually in four, since there is `generic-whitelist` + `jenkins-whitelist`. OTOH all the classes in `generic-whitelist` are resolvable except for one `groovy-cps` thing which has been obsolete for a long time and could probably just be deleted now.

PR #259, opened 2019-06-04
[JENKINS-56682] Fix use of @Field in sandboxed scripts during GroovyShell.parse

dwnusbaum:
See [JENKINS-56682](https://issues.jenkins-ci.org/browse/JENKINS-56682).

The fix for SECURITY-1336 broke the use of variables annotated with `@Field` in sandboxed scripts where those variables referenced other `@Field` variables in their initializer.

This PR attempts to fix the issue by making `SandboxInterceptor` check for declared fields of the Script subclass, not just bindings, and return their values when appropriate.

The fix here only affects the use of `SecureGroovyScript` and other callers of `GroovySandbox.runScript`, it does not affect Pipeline.


src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScriptTest.java
```diff
@@ -993,4 +993,18 @@ public void blockConstructorInvocationAtRuntime() throws Exception {
...
+    public void testScriptFieldsAddedViaAtField() throws Exception {
+        FreeStyleProject p = r.createFreeStyleProject();
+        p.getPublishersList().add(new TestGroovyRecorder(new SecureGroovyScript(
+                "import groovy.transform.Field\n" +
+                "@Field foo = 1\n" +
+                "@Field bar = foo + 1\n" + // evaluated during GroovyShell.parse
+                "if (bar != 2) {\n" +
+                "  throw new Exception('oops')\n" +
+                "}\n", true, null)));
+        r.buildAndAssertSuccess(p);
```
dwnusbaum: Previous stack trace:

```
groovy.lang.MissingPropertyException: No such property: foo for class: groovy.lang.Binding
    at groovy.lang.Binding.getVariable(Binding.java:63)
    at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptor.onGetProperty(SandboxInterceptor.java:270)
    at org.kohsuke.groovy.sandbox.impl.Checker$6.call(Checker.java:289)
    at org.kohsuke.groovy.sandbox.impl.Checker.checkedGetProperty(Checker.java:293)
    at org.kohsuke.groovy.sandbox.impl.Checker$checkedGetProperty.callStatic(Unknown Source)
    at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCallStatic(CallSiteArray.java:56)
    at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callStatic(AbstractCallSite.java:194)
    at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callStatic(AbstractCallSite.java:230)
    at Script1.<init>(Script1.groovy:3)
Caused: java.lang.reflect.InvocationTargetException
    at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
    at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
    at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
    at org.codehaus.groovy.runtime.InvokerHelper.createScript(InvokerHelper.java:431)
Caused: groovy.lang.GroovyRuntimeException: Failed to create Script instance for class: class Script1. Reason
    at org.codehaus.groovy.runtime.InvokerHelper.createScript(InvokerHelper.java:466)
    at groovy.lang.GroovyShell.parse(GroovyShell.java:700)
    at groovy.lang.GroovyShell.parse(GroovyShell.java:736)
    at groovy.lang.GroovyShell.parse(GroovyShell.java:727)
    at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.GroovySandbox.runScript(GroovySandbox.java:159)
    at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SecureGroovyScript.evaluate(SecureGroovyScript.java:345)
    at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.TestGroovyRecorder.perform(TestGroovyRecorder.java:62)
    ...
```

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -265,11 +265,19 @@
 
     @Override public Object onGetProperty(GroovyInterceptor.Invoker invoker, final Object receiver, final String property) throws Throwable {
         MissingPropertyException mpe = null;
-        if (receiver instanceof Script) { // SimpleTemplateEngine "out" variable, and anything else added in a binding
+        if (receiver instanceof Script) { // SimpleTemplateEngine "out" variable, anything else added in a binding, and @Field variables.
             try {
                 ((Script) receiver).getBinding().getVariable(property); // do not let it go to Script.super.getProperty
                 return super.onGetProperty(invoker, receiver, property);
             } catch (MissingPropertyException x) {
+                if (receiver.getClass() != Script.class) { // Only check declared fields of subclasses, not Script itself
```
dwnusbaum: Maybe unnecessary, since binding is the only field declared directly in `Script`.

dwnusbaum:
> We are already checking GroovyCallSiteSelector.field below, and if that passes it ought to ignore mpe. So why is that not passing

From my debugging, it looks like it finds the field, but the field isn't whitelisted, so [this condition is false](https://github.com/jenkinsci/script-security-plugin/blob/294a80c190adefeace621ff73c74c6d956b403e0/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java#L339). I experimented with creating a dynamic `EnumeratingWhitelist` containing the fields of the script, but because this happens during `GroovyShell.parse`, we don't the script in advance, so I'm not sure where or how we'd update the whitelist after `FieldASTTransformation` runs.

The strange thing to me is that uses of `@Field` variables seem to work correctly at runtime, just not during compilation. I'm not sure why that is happening.

jglick:
> it finds the field, but the field isn't whitelisted

Because [this](https://github.com/jenkinsci/script-security-plugin/blob/294a80c190adefeace621ff73c74c6d956b403e0/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java#L164) is not yet in effect. Maybe try something like

```diff
diff --git a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java
index 8028f4d..a397aef 100644
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java
@@ -154,16 +154,12 @@ public final class GroovySandbox {
      * @return the return value of the script
      */
     public Object runScript(@Nonnull GroovyShell shell, @Nonnull String script) {
-        Script s;
-        try (Scope scope = enter()) {
-            s = shell.parse(script);
-        }
         GroovySandbox derived = new GroovySandbox().
             withApprovalContext(context).
             withTaskListener(listener).
-            withWhitelist(new ProxyWhitelist(new ClassLoaderWhitelist(s.getClass().getClassLoader()), whitelist()));
+            withWhitelist(new ProxyWhitelist(new ClassLoaderWhitelist(shell.getClassLoader()), whitelist()));
         try (Scope scope = derived.enter()) {
-            return s.run();
+            return shell.parse(script).run();
         }
     }
 
```

dwnusbaum:
That breaks a lot of tests that use methods and variables defined on the class during runtime, but does fix the compile-time issue. Along those lines, the following seems to fix the reported issue without breaking existing tests (although I think we'd want to actually create a new sandbox rather than calling `withWhitelist` directly):

```diff
diff --git a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java
index 8028f4d..155f712 100644
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java
@@ -155,6 +155,7 @@ public final class GroovySandbox {
      */
     public Object runScript(@Nonnull GroovyShell shell, @Nonnull String script) {
         Script s;
+        withWhitelist(new ProxyWhitelist(new ClassLoaderWhitelist(shell.getClassLoader()), whitelist()));
         try (Scope scope = enter()) {
             s = shell.parse(script);
         }
```

jglick:
> we'd want to actually create a new sandbox rather than calling `withWhitelist` directly

Yes.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -865,8 +866,18 @@ public void blockedASTTransformsGrab() throws Exception {
...
-        GroovyShell shell = new GroovyShell(GroovySandbox.createSecureCompilerConfiguration());
-        Object actual = GroovySandbox.run(shell, script, whitelist);
+        CompilerConfiguration cc = GroovySandbox.createSecureCompilerConfiguration();
+        GroovyShell shell = new GroovyShell(cc);
+        try {
+            Field loaderF = GroovyShell.class.getDeclaredField("loader");
+            loaderF.setAccessible(true);
+            ClassLoader loader = GroovyShell.class.getClassLoader();
+            ClassLoader memoryProtectedLoader = new SecureGroovyScript.CleanGroovyClassLoader(GroovySandbox.createSecureClassLoader(loader), cc);
+            loaderF.set(shell, memoryProtectedLoader);
```
dwnusbaum: It turns out that CleanGroovyClassLoader _must_ be the loader used by the shell (and in turn also the script) for fields to be whitelisted correctly in initializers. I think the reason is [what is mentioned in this line](https://github.com/jenkinsci/script-security-plugin/blob/294a80c190adefeace621ff73c74c6d956b403e0/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java#L417) in CleanGroovyClassLoader's Javadoc. If you do not use this loader, then the new tests in `SandboxInterceptor` fail. I think that means we'll need some kind of similar change [here](https://github.com/jenkinsci/workflow-cps-plugin/blob/d81da7167828987d3d8d41d52617acfbee86705b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShellFactory.java#L114) in workflow-cps to modify the class loader used in Pipeline? Either way it will definitely need to be tested.

dwnusbaum: Ok, workflow-cps already has a copy of `CleanGroovyClassLoader` it uses for `CpsGroovyShell`, so the fix on that side is just to make sure that it adds a `ClassLoaderWhitelist ` for the shell's class loader when parsing the script. See https://github.com/jenkinsci/workflow-cps-plugin/pull/297.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovySandbox.java
```diff
@@ -154,16 +154,12 @@ public Scope enter() {
...
     public Object runScript(@Nonnull GroovyShell shell, @Nonnull String script) {
-        Script s;
-        try (Scope scope = enter()) {
-            s = shell.parse(script);
-        }
         GroovySandbox derived = new GroovySandbox().
             withApprovalContext(context).
             withTaskListener(listener).
-            withWhitelist(new ProxyWhitelist(new ClassLoaderWhitelist(s.getClass().getClassLoader()), whitelist()));
+            withWhitelist(new ProxyWhitelist(new ClassLoaderWhitelist(shell.getClassLoader()), whitelist()));
```
dwnusbaum: A bit awkward, because this only fixes the issue if the caller has modified `shell.loader` reflectively to make it something like `CleanGroovyClassLoader`, but we don't have a good way to enforce that condition. Should we provide a `GroovySandbox.prepareGroovyShell` method as well that sets up `CleanGroovyClassLoader` for other callers?

dwnusbaum: Also if there are any direct callers doing something like SandboxInterceptorTest that aren't using the `CleanGroovyClassLoader` trick, then I think this is a breaking change that will block any types defined in the script because `shell.getClassloader()` will be different than `s.getClass().getClassLoader()`. That said, I don't see any existing callers of `runScript` outside of this plugin.

jglick: > this only fixes the issue

I.e., the regression. So long as a misuse of this method does not result in a _vulnerability_, I think it is good enough.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -865,8 +866,18 @@ public void blockedASTTransformsGrab() throws Exception {
     }
 
     private static Object evaluate(Whitelist whitelist, String script) {
-        GroovyShell shell = new GroovyShell(GroovySandbox.createSecureCompilerConfiguration());
-        Object actual = GroovySandbox.run(shell, script, whitelist);
+        CompilerConfiguration cc = GroovySandbox.createSecureCompilerConfiguration();
+        GroovyShell shell = new GroovyShell(cc);
+        try {
+            Field loaderF = GroovyShell.class.getDeclaredField("loader");
```
jglick: Is there a way some of the subtle code here can be shared with `SecureGroovyScript.evaluate`? Is that what you were proposing in https://github.com/jenkinsci/script-security-plugin/pull/259#discussion_r294446731?

dwnusbaum: Yeah, I think we could create something like `static void setGroovyClassLoader(GroovyShell, ClassLoader)` that sets the loader field after wrapping the passed loader in CleanGroovyClassLoader, and then update both places to call that method. 

Maybe possible to go further and introduce a new subtype of `GroovyShell` that does that in the constructor so that we could also remove the duplication in [`CpsGroovyShell`](https://github.com/jenkinsci/workflow-cps-plugin/blob/7eab5dc0c39289a372a05904d65596c4192c4005/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java#L50-L57) (might need to tweak the memory leak fixes in workflow-cps if we did do that). We could also add a `GroovyShell prepareSecureGroovyShell(ClassLoader, Binding, CompilerConfiguration)` in `GroovySandbox` and recommend downstream users of the plugin use that instead of instantiating `GroovyShell` directly.

jglick: Could all be deferred to some follow-up. Maybe capture the conversation in a TODO comment or something.

dwnusbaum: TODO added in 9e8a1ae.

PR #260, opened 2019-07-23
Update whitelist DefaultGroovyMethods {any, every}

pitkley:
Like for `each`, `any` and `every` might receive a `java.lang.Object` although the object itself is in fact iterable.

Minified example:

```groovy
'string'.split().every { it }
```

-----

I'm not sure if this introduces any security-risks, and if there is a way to verify this. Should there be anything else necessary (like tests), please let me know!


bitwiseman:
@dwnusbaum @jglick 
Do you have a process for evaluating these requests?  Are there specific places to look for security vulnerability reports?

jglick:
When in doubt, just check the source code. Does it do anything weird or scary? Not very scientific, I know.

dwnusbaum:
> Are there specific places to look for security vulnerability reports?

No, there isn't anything along those lines that would be relevant for our use case as far as I am aware. The [blacklist](https://github.com/jenkinsci/script-security-plugin/blob/master/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/blacklist) might help give you an idea of things to look out for, generally things like metaprogramming, File-related methods, and access to Jenkins APIs.

dwnusbaum:
Actually, I want to run a quick test.

dwnusbaum:
Test was fine, I was overthinking things and worried about the methods calling `$object.iterator` using reflection, but any implementation of `iterator` is still transformed by the sandbox so there are no issues.

pitkley:
@dwnusbaum thanks for testing and merging! :+1:

PR #261, opened 2019-08-04
JENKINS-58799: Whitelisting a few collection related methods

haridsv:
This PR whitelists mainly a few collection methods that will enable access using range. It also includes a few Groovy collection enhancements in the list.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -32,6 +32,12 @@ method groovy.lang.Closure ncurry int java.lang.Object[]
 method groovy.lang.Closure setDelegate java.lang.Object
 method groovy.lang.Closure setResolveStrategy int
 method groovy.lang.GString plus java.lang.String
+method groovy.lang.NumberRange by int
+method groovy.lang.NumberRange getStepSize
+method groovy.lang.Range getFrom
+method groovy.lang.Range getTo
+method groovy.lang.Range step int
+method groovy.lang.Range step int groovy.lang.Closure
 method groovy.lang.Script getBinding
 
 # Useful to show stacktrace to the user.
@@ -332,6 +338,7 @@ staticField java.util.Locale UK
 staticField java.util.Locale US
 method java.util.Map clear
 method java.util.Map containsKey java.lang.Object
+method java.util.Map containsValue java.lang.Object
 method java.util.Map entrySet
 method java.util.Map get java.lang.Object
 method java.util.Map getOrDefault java.lang.Object java.lang.Object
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -577,6 +585,8 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.Co
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.List java.lang.Iterable
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.List java.lang.Object
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.List java.util.Collection
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.List removeAt int
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.List removeLast
```
dwnusbaum: @haridsv These entries are malformed. `removeAt` and `removeLast` aren't types, what methods were you trying to whitelist? See https://github.com/jenkinsci/script-security-plugin/pull/268.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -32,6 +32,12 @@ method groovy.lang.Closure ncurry int java.lang.Object[]
 method groovy.lang.Closure setDelegate java.lang.Object
 method groovy.lang.Closure setResolveStrategy int
 method groovy.lang.GString plus java.lang.String
+method groovy.lang.NumberRange by int
+method groovy.lang.NumberRange getStepSize
```
dwnusbaum: @haridsv `NumberRange` does not exist in Groovy 2.4.x as far as I can tell, which is what Jenkins is currently using, so these entries don't make sense. See https://github.com/jenkinsci/script-security-plugin/pull/268.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
haridsv: I think my intention was to whitelist these two methods: http://docs.groovy-lang.org/latest/html/groovy-jdk/java/util/List.html#removeAt(int) and http://docs.groovy-lang.org/latest/html/groovy-jdk/java/util/List.html#removeLast() so I seem to have started with wrong lines and not realize because there was no test failure :(

I think the right entries would have been:

```
staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods removeAll java.util.Collection groovy.lang.Closure
staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods removeLast java.util.List
```

However, the last one is only available in Groovy 2.5.0, so can't work. In case you don't mind taking this change forward via you PR, I will add a suggestion in your other RP.

PR #262, opened 2019-08-12
JENKINS-58894 Add String signatures to match CharSequence counterparts

hogarthj:
This adds a variety of java.lang.String methods from StringGroovyMethods to match the CharSequence versions.

This also adds readLines as an acceptable method.


hogarthj:
@bitwiseman split out as requested :)

PR #265, opened 2019-09-23
[JENKINS-59587] Use weakValues on the inner caches in SandboxResolvingClassLoader.parentClassCache instead of on the outer cache

dwnusbaum:
See [JENKINS-59587](https://issues.jenkins-ci.org/browse/JENKINS-59587).

While investigating a user-reported performance issue, I noticed among other things that their Pipelines were spending a ton of time blocked waiting to acquire the lock for `PluginManager$UberClassLoader` while loading classes in `SandboxResolvingClassLoader`. This was a little surprising, because this user has a single shared library that is used by all of their Pipeline builds, so I would've thought that they would have a very high cache hit ratio, but that did not seem to be the case.

Looking into it, I noticed that in `SandboxResolvingClassLoader`, we were using `weakValues` on `parentClassCache` itself as of #252. This didn't make sense to me, because the values of `parentClassCache` are `Cache<String, Class<?>>`, and from what I can tell, these caches are only referenced internally by `parentClassCache`, which meant that the inner caches would always be removed when the garbage collector runs, making the cache not very useful when Jenkins is under significant memory pressure (which was definitely case for this user for other reasons).

This PR switches to using `weakValues` on `parentClassCache`'s inner caches, rather than `parentClassCache` itself. This way, the classes referenced by the inner cache are still only weakly referenced, so that their class loader, which may itself be a key of `parentClassCache`, is still able to be garbage collected. Looking into it some more, I think #252 was doing this already before 36d3b3a2a5b8561079cdccde53db85000dd40979, but that commit changed the semantics from `Cache<ClassLoader, Cache<String, Optional<WeakReference<Class<?>>>>>` to `Cache<ClassLoader, WeakReference<Cache<String, Optional<Class<?>>>>>`, and unfortunately we didn't notice that change in semantics during review.

I added a basic smoke test that fails if you revert the changes in `src/main`. `GroovyMemoryLeakTest` still passes, but I'm not sure what other testing would be useful to make sure this doesn't cause any regressions.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
@@ -23,9 +23,9 @@ class SandboxResolvingClassLoader extends ClassLoader {
 
     private static final Logger LOGGER = Logger.getLogger(SandboxResolvingClassLoader.class.getName());
 
-    private static final LoadingCache<ClassLoader, Cache<String, Class<?>>> parentClassCache = makeParentCache(true);
+    static final LoadingCache<ClassLoader, Cache<String, Class<?>>> parentClassCache = makeParentCache(true);
 
-    private static final LoadingCache<ClassLoader, Cache<String, Optional<URL>>> parentResourceCache = makeParentCache(false);
+    static final LoadingCache<ClassLoader, Cache<String, Optional<URL>>> parentResourceCache = makeParentCache(false);
 
     SandboxResolvingClassLoader(ClassLoader parent) {
         super(parent);
@@ -38,7 +38,7 @@ class SandboxResolvingClassLoader extends ClassLoader {
      * This value is non-null, not a legitimate return value
      * (no script should be trying to load this implementation detail), and strongly held.
      */
-    private static final Class<?> CLASS_NOT_FOUND = Unused.class;
+    static final Class<?> CLASS_NOT_FOUND = Unused.class;
     private static final class Unused {}
 
     @Override protected synchronized Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
@@ -93,12 +93,18 @@ class SandboxResolvingClassLoader extends ClassLoader {
         });
     }
 
-    private static <T> LoadingCache<ClassLoader, Cache<String, T>> makeParentCache(boolean weakValues) {
-        Caffeine<Object, Object> builder = Caffeine.newBuilder().weakKeys();
-        if (weakValues) {
-            builder = builder.weakValues();
+    private static <T> LoadingCache<ClassLoader, Cache<String, T>> makeParentCache(boolean weakValuesInnerCache) {
+        // The outer cache has weak keys, so that we do not leak class loaders, but strong values, because the
+        // inner caches are only referenced by the outer cache internally.
+        Caffeine<Object, Object> outerBuilder = Caffeine.newBuilder().weakKeys();
+        // The inner cache has strong keys, since they are just strings, and expires entries 15 minutes after they are
+        // added to the cache (TODO: should we use expireAfterAccess instead?). The values for the inner cache may
+        // be weak if needed, for example parentClassCache uses weak values to avoid leaking classes and their loaders.
+        Caffeine<Object, Object> innerBuilder = Caffeine.newBuilder().expireAfterWrite(Duration.ofMinutes(15));
+        if (weakValuesInnerCache) {
+            innerBuilder.weakValues();
         }
 
-        return builder.build(parentLoader -> Caffeine.newBuilder()./* allow new plugins to be used, and clean up memory */expireAfterWrite(Duration.ofMinutes(15)).build());
+        return outerBuilder.build(parentLoader -> innerBuilder.build());
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
@@ -98,8 +98,10 @@ class SandboxResolvingClassLoader extends ClassLoader {
         // inner caches are only referenced by the outer cache internally.
         Caffeine<Object, Object> outerBuilder = Caffeine.newBuilder().weakKeys();
         // The inner cache has strong keys, since they are just strings, and expires entries 15 minutes after they are
-        // added to the cache (TODO: should we use expireAfterAccess instead?). The values for the inner cache may
-        // be weak if needed, for example parentClassCache uses weak values to avoid leaking classes and their loaders.
+        // added to the cache, so that classes defined by dynamically installed plugins become available even if there
+        // were negative cache hits prior to the installation (ideally this would be done with a listener). The values
+        // for the inner cache may be weak if needed, for example parentClassCache uses weak values to avoid leaking
+        // classes and their loaders.
         Caffeine<Object, Object> innerBuilder = Caffeine.newBuilder().expireAfterWrite(Duration.ofMinutes(15));
         if (weakValuesInnerCache) {
             innerBuilder.weakValues();
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
@@ -96,13 +96,13 @@ class SandboxResolvingClassLoader extends ClassLoader {
     private static <T> LoadingCache<ClassLoader, Cache<String, T>> makeParentCache(boolean weakValuesInnerCache) {
         // The outer cache has weak keys, so that we do not leak class loaders, but strong values, because the
         // inner caches are only referenced by the outer cache internally.
-        Caffeine<Object, Object> outerBuilder = Caffeine.newBuilder().weakKeys();
+        Caffeine<Object, Object> outerBuilder = Caffeine.newBuilder().recordStats().weakKeys();
         // The inner cache has strong keys, since they are just strings, and expires entries 15 minutes after they are
         // added to the cache, so that classes defined by dynamically installed plugins become available even if there
         // were negative cache hits prior to the installation (ideally this would be done with a listener). The values
         // for the inner cache may be weak if needed, for example parentClassCache uses weak values to avoid leaking
         // classes and their loaders.
-        Caffeine<Object, Object> innerBuilder = Caffeine.newBuilder().expireAfterWrite(Duration.ofMinutes(15));
+        Caffeine<Object, Object> innerBuilder = Caffeine.newBuilder().recordStats().expireAfterWrite(Duration.ofMinutes(15));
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
```diff
@@ -93,12 +93,18 @@
...
+        // The inner cache has strong keys, since they are just strings, and expires entries 15 minutes after they are
+        // added to the cache (TODO: should we use expireAfterAccess instead?). The values for the inner cache may
+        // be weak if needed, for example parentClassCache uses weak values to avoid leaking classes and their loaders.
+        Caffeine<Object, Object> innerBuilder = Caffeine.newBuilder().expireAfterWrite(Duration.ofMinutes(15));
+        if (weakValuesInnerCache) {
+            innerBuilder.weakValues();
         }
 
-        return builder.build(parentLoader -> Caffeine.newBuilder()./* allow new plugins to be used, and clean up memory */expireAfterWrite(Duration.ofMinutes(15)).build());
+        return outerBuilder.build(parentLoader -> innerBuilder.build());
```
dwnusbaum: Note that reusing `innerBuilder.build` should be safe according to [its Javadoc](https://static.javadoc.io/com.github.ben-manes.caffeine/caffeine/2.8.0/com/github/benmanes/caffeine/cache/Caffeine.html#build--).

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
```diff
@@ -93,12 +93,18 @@
...
+        Caffeine<Object, Object> outerBuilder = Caffeine.newBuilder().weakKeys();
+        // The inner cache has strong keys, since they are just strings, and expires entries 15 minutes after they are
+        // added to the cache (TODO: should we use expireAfterAccess instead?). The values for the inner cache may
+        // be weak if needed, for example parentClassCache uses weak values to avoid leaking classes and their loaders.
+        Caffeine<Object, Object> innerBuilder = Caffeine.newBuilder().expireAfterWrite(Duration.ofMinutes(15));
+        if (weakValuesInnerCache) {
+            innerBuilder.weakValues();
         }
 
-        return builder.build(parentLoader -> Caffeine.newBuilder()./* allow new plugins to be used, and clean up memory */expireAfterWrite(Duration.ofMinutes(15)).build());
```
jglick: Might want to retain a mention in comments that it is important to expire entries _somehow_ in order to let new plugins be used. Ideally this would be done with a listener.

dwnusbaum: Ah ok, it wasn't totally clear to me what that comment was referring to. I will add it back.

dwnusbaum:
One other thing that seemed a bit concerning in the thread dumps I saw was that some class loading threads held locks on a `ConcurrentHashMap$Node` while they were waiting for the lock for `PluginManager$UberClassLoader`, and other threads were then blocked waiting for that same `ConcurrentHashMap$Node`. Caffeine's [`Cache.get` Javadoc](https://static.javadoc.io/com.github.ben-manes.caffeine/caffeine/2.8.0/com/github/benmanes/caffeine/cache/Cache.html#get-K-java.util.function.Function-) states that "Some attempted update operations on this cache by other threads may be blocked while the computation is in progress, so the computation should be short and simple, and must not attempt to update any other mappings of this cache". 

Looking into the code, I think [`BoundedLocalCache.computeIfAbsent`](https://github.com/ben-manes/caffeine/blob/19f0911a78fa8096a4f610a24b756741dd4ed984/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java#L2335-L2361) first tries to do a lookup without locking, and the locking only occurs if the value needs to be computed. In our case, I don't think the inner cache can load more than value concurrently from `PluginManager$UberClassLoader` anyway, so I don't know if this actually matters. IIUC, the preferred approach for slow mapping functions is to use `AsyncCache`, but it seems like in our case that would just move some of the locking from `ConcurrentHashMap$Node` to `PluginManager$UberClassLoader` for cache misses while cache hits would be unaffected because of the no-lock happy path.

Example thread holding a `ConcurrentHashMap$Node` lock:

```
java.lang.Thread.State: BLOCKED (on object monitor)
  at java.lang.ClassLoader.loadClass(ClassLoader.java:404)
  - waiting to lock <0x00000003c21e35f0> (a hudson.PluginManager$UberClassLoader)
  at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
  at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxResolvingClassLoader.lambda$loadClass$0(SandboxResolvingClassLoader.java:51)
  at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxResolvingClassLoader$$Lambda$269/265156624.get(Unknown Source)
  at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxResolvingClassLoader.lambda$load$2(SandboxResolvingClassLoader.java:85)
  at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxResolvingClassLoader$$Lambda$271/201356224.apply(Unknown Source)
  at com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$14(BoundedLocalCache.java:2337)
  at com.github.benmanes.caffeine.cache.BoundedLocalCache$$Lambda$270/1614184557.apply(Unknown Source)
  at java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1892)
  - locked <0x000000078a3ed640> (a java.util.concurrent.ConcurrentHashMap$Node)
  at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2335)
  at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2318)
  at com.github.benmanes.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:111)
  at com.github.benmanes.caffeine.cache.LocalManualCache.get(LocalManualCache.java:54)
  at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxResolvingClassLoader.load(SandboxResolvingClassLoader.java:79)
```

Example thread blocked waiting for that `ConcurrentHashMap$Node`:

```
java.lang.Thread.State: BLOCKED (on object monitor)
  at java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1868)
  - waiting to lock <0x000000078a3ed640> (a java.util.concurrent.ConcurrentHashMap$Node)
  at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2335)
  at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2318)
  at com.github.benmanes.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:111)
  at com.github.benmanes.caffeine.cache.LocalManualCache.get(LocalManualCache.java:54)
  at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxResolvingClassLoader.load(SandboxResolvingClassLoader.java:79)
```

dwnusbaum:
I enabled statistics collection so that we can easily evaluate cache performance through the script console with a snippet like this:

```
import com.github.benmanes.caffeine.cache.stats.CacheStats;
import org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxResolvingClassLoader;
ClassLoader loader = Jenkins.getInstance().getPluginManager().uberClassLoader;
CacheStats stats = SandboxResolvingClassLoader.parentClassCache.get(loader).stats();
println(stats.hitRate());
```

jglick:
```
java.lang.AssertionError: 

Expected: <2L>
     but: was <795L>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.junit.Assert.assertThat(Assert.java:923)
	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxResolvingClassLoaderTest.classCacheDoesNotHoldClassValuesTooWeakly(SandboxResolvingClassLoaderTest.java:64)
```

jglick:
> evaluate cache performance through the script console

Can this be done via `metrics`, or `support-core`, or something?

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoaderTest.java
```diff
@@ -0,0 +1,67 @@
...
+        assertThat(parentClassCache.get(parentLoader).getIfPresent("this.does.not.Exist"), equalTo(CLASS_NOT_FOUND));
+        // Make sure that both of the entries are still in the cache after a GC.
+        System.gc();
+        assertThat(parentClassCache.get(parentLoader).getIfPresent("java.lang.String"), equalTo(String.class));
+        assertThat(parentClassCache.get(parentLoader).getIfPresent("this.does.not.Exist"), equalTo(CLASS_NOT_FOUND));
+        CacheStats stats = parentClassCache.get(parentLoader).stats();
+        // Before the fix for JENKINS-59587, the original inner cache was removed after the call to `System.gc()`, so
+        // the miss count was 2 and the hit count was 0.
+        assertThat(stats.missCount(), equalTo(2L)); // The two calls to `loadClass()`
+        assertThat(stats.hitCount(), equalTo(4L)); // The four calls to `getIfPresent()`
```
jglick: fails

dwnusbaum: Bizarre, it passes locally. I'll look into it.

jglick:
```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <configuration>
        <reuseForks>false</reuseForks>
    </configuration>
</plugin>
```

I guess we should just make this the default in `plugin-pom` already.

dwnusbaum:
> Can this be done via metrics, or support-core, or something?

Sure, now that we are recording stats, we could hook them up to whatever extension points might use them, although I'd prefer not to add any dependencies to this plugin if possible.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoaderTest.java
dwnusbaum: Thanks for the pointer, fixed in 0a222ce.

dwnusbaum:
I explored the things mentioned in https://github.com/jenkinsci/script-security-plugin/pull/265#issuecomment-534251537 in https://github.com/jenkinsci/script-security-plugin/pull/271. In summary, unless [JENKINS-23784](https://issues.jenkins-ci.org/browse/JENKINS-23784) was addressed, using an asynchronous cache makes no difference, so I think it makes sense to keep using the manual cache for now.

PR #268, opened 2019-10-01
Remove useless whitelist entries, fix typo, and improve tests

dwnusbaum:
#249 and #261 had some problems. `NumberRange` does not exist in Groovy 2.4.x, which is what Jenkins is using, the `DefaultGroovyMethods.minus` entries are malformed because of `removeAt` and `removeLast`, and `Flushable` is in the `java.io` package, not `java.lang`.

CC @haridsv

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -32,14 +32,13 @@ method groovy.lang.Closure ncurry int java.lang.Object[]
 method groovy.lang.Closure setDelegate java.lang.Object
 method groovy.lang.Closure setResolveStrategy int
 method groovy.lang.GString plus java.lang.String
-method groovy.lang.NumberRange by int
-method groovy.lang.NumberRange getStepSize
 method groovy.lang.Range getFrom
 method groovy.lang.Range getTo
 method groovy.lang.Range step int
 method groovy.lang.Range step int groovy.lang.Closure
 method groovy.lang.Script getBinding
 
+method java.io.Flushable flush
 # Useful to show stacktrace to the user.
 new java.io.PrintWriter java.io.Writer
 # Needed for Groovy Templates to emit output:
@@ -80,7 +79,6 @@ new java.lang.Enum java.lang.String int
 method java.lang.Enum name
 method java.lang.Enum ordinal
 new java.lang.Exception java.lang.String
-method java.lang.Flushable flush
 staticField java.lang.Integer MAX_VALUE
 # could add valueOf, though currently the staticField’s need to be whitelisted, which is the more likely use case
 staticMethod java.lang.Integer parseInt java.lang.String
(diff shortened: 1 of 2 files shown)


haridsv:
Thanks @dwnusbaum !

> NumberRange does not exist in Groovy 2.4.x

Got it! Does this mean that the unit tests don't use the same version of Groovy to have not caught this?

> DefaultGroovyMethods.minus entries are malformed because of removeAt and removeLast

Just so I understand what is the exact issue here? Why won't the unit tests catch this?

> Flushable is in the java.io package, not java.lang

Indeed! I am not sure how I got that mixed up, but then again, why did the unit tests pass?

haridsv:
Just saw your comment on the older PR about the `minus`, will respond there.

haridsv:
I just ran the tests again with the old changes and while there is no test failure, I noticed the below messages:

```
Cannot check validity of `method groovy.lang.NumberRange by int` due to java.lang.ClassNotFoundException: groovy.lang.NumberRange
Cannot check validity of `method groovy.lang.NumberRange getStepSize` due to java.lang.ClassNotFoundException: groovy.lang.NumberRange
Cannot check validity of `method java.lang.Flushable flush` due to java.lang.ClassNotFoundException: java.lang.Flushable
```

I also noticed this additional message:

```
Cannot check validity of `staticMethod com.cloudbees.groovy.cps.CpsDefaultGroovyMethods each java.util.Iterator groovy.lang.Closure` due to java.lang.ClassNotFoundException: com.cloudbees.groovy.cps.CpsDefaultGroovyMethods
```

Looks like from the PR #191 (FYI @abayer)

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -587,8 +585,6 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.Co
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.List java.lang.Iterable
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.List java.lang.Object
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.List java.util.Collection
-staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.List removeAt int
```
haridsv: Would you mind adding the right entry for it? It should be:

```
staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods removeAll java.util.Collection groovy.lang.Closure
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -587,8 +585,6 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.Co
...
-staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods minus java.util.List removeLast
```
haridsv: This is only applicable since Groovy 2.5.0, but the correct entry would have been:

```
staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods removeLast java.util.List
```

dwnusbaum:
@haridsv The reason the tests did not fail is that a failing signature check is not fatal, specifically because of cases like this:

> Cannot check validity of `staticMethod com.cloudbees.groovy.cps.CpsDefaultGroovyMethods each java.util.Iterator groovy.lang.Closure` due to java.lang.ClassNotFoundException: com.cloudbees.groovy.cps.CpsDefaultGroovyMethods

That is a real method signature from groovy-cps, important for use of the sandbox during Pipeline execution, but that library is not on the classpath during these tests, so it cannot be found. Maybe we should special case it and any other cases like it, and then make a non-existent signature fail the build to catch these kinds of issues automatically?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
dwnusbaum: It looks like that method is already whitelisted: https://github.com/jenkinsci/script-security-plugin/blob/54e05e650caaf53231f0df908b78c9b83110e52b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist#L726

PR #271, opened 2019-10-08
[PoC] Switch from manually loaded to asynchronously loaded inner caches

dwnusbaum:
While working on https://github.com/jenkinsci/script-security-plugin/pull/265, I noticed that in some cases, class loading threads were blocked on internal `ConcurrentHashMap` data structures locked by other class loading threads, rather than on their parent class loader. See https://github.com/jenkinsci/script-security-plugin/pull/265#issuecomment-534251537 for details. I thought that switching to asynchronously-loaded inner caches might help this problem, and it does, but in the end, it doesn't really matter because `PluginManager.UberClassLoader` is not parallel capable ([JENKINS-23784](https://issues.jenkins-ci.org/browse/JENKINS-23784)), and as I thought, cache hits are fine either way because they do not require any locking.

I am opening this as a draft PR, and then closing it, because it adds some complexity that I'd prefer to avoid unless it actually helps us, and it needs some more testing and thought around error handling to be ready to merge.

See `SandboxResolvingClassLoaderTest.classCacheOnlyBlocksAtLoadClass` for the relevant test that fails without this change. The main benefit from this change is that the values for inner caches become `Future`s, so the insertion of the `Future` into the cache is quick, and the `ConcurrentHashMap`-internal locks are not held during the duration of calls to `parentLoader.loadClass`. The results of the `Future` are computed on a separate thread pool. If `PluginManager.UberClassLoader` was parallel capable, this change would allow multiple classes to be loaded simultaneously from the thread pool. If `PluginManager.UberClassLoader` was parallel capable but we were still using the manually-loaded cache, class loading would in practice still be serialized on internal `ConcurrentHashMap` data structures, which would be locked during the duration of calls to `parentLoader.loadClass`.

TL;DR: Without [JENKINS-23784](https://issues.jenkins-ci.org/browse/JENKINS-23784), using an async cache makes no difference to performance.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxResolvingClassLoader.java
@@ -1,12 +1,15 @@
 package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
 
-import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.AsyncCache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import groovy.lang.GroovyShell;
+import java.lang.ref.WeakReference;
 import java.net.URL;
 import java.time.Duration;
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


ben-manes:
> class loading would in practice still be serialized on internal ConcurrentHashMap data structures,

fyi, in that case you would probably prefer setting a higher `initialCapacity` to increase the number of locks in `ConcurrentHashMap`. The count is based on the hash table's size, where entries belonging to the same bin share a lock. If a cache stays small (e.g. aggressively expires) then the cache won't grow to have many locks, which could impact the write performance. If you know a high write rate is needed for a small cache, then setting a higher initial capacity is helpful. It's not as perfectly parallel as `AsyncCache`, but is simpler code especially given the `WeakReference` requirement.

dwnusbaum:
Thanks for taking a look!

> fyi, in that case you would probably prefer setting a higher initialCapacity to increase the number of locks in ConcurrentHashMap. The count is based on the hash table's size, where entries belonging to the same bin share a lock. 

Good point, I expect the cache to have a relatively large number of entries, given it is shared across the entire Jenkins instance, so starting with a higher initial capacity is probably a good idea.

> If a cache stays small (e.g. aggressively expires) then the cache won't grow to have many locks, which could impact the write performance.

Good point about expiration. Right now, we just expire everything every 15 minutes, but that is not as specific as it could be. Ideally, positive cache hits (loaded classes in our case) would never expire, and negative cache hits (class names not in any of the parent class loaders in our case) would only expire when a new plugin is installed in Jenkins without the JVM being restarted (which is a relatively rare occurrence, since plugin upgrades always require a restart). This is probably the best place to look to for further improvements in this area, as the Groovy compiler seems to do a bunch of speculative class loading to understand how to interpret some pieces of syntax, so expiring the negative hits less frequently would avoid spikes of class loading activity.

ben-manes:
> Ideally, positive cache hits (loaded classes in our case) would never expire, and negative cache hits (class names not correspond to a real class according to any of the parent class loaders in our case) would only expire when a new plugin is installed

You might consider using a custom `expireAfter(Expiry)` that uses a different duration based on a positive vs negative cache load (`CLASS_NOT_FOUND` result). This way you give a longer TTL for positive hits. We don't yet have, though requested, a similar variant of `refreshAfterWrite` (e.g. `refreshAfter(refresher)`) as that could be of interest for you too.

PR #274, opened 2019-10-27
JENKINS-59952: Whitelisting all remaining java.util.regex methods

haridsv:
This change ensures that all the methods from java.util.regex package are whitelisted. There is no functionality in this package that touches external resources, so everything should be safe to be whitelisted.


haridsv:
I only observed those that are already known, so yes, none of these that I am adding showed up there.

ketan:
Could someone please kick off the build again? It seems there was some error with the build.

haridsv:
Yes, there seemed to be a transient issue, but couldn't find an option on the Jenkins interface. I could try pushing an empty commit.

dwnusbaum:
Very few users have permission to trigger a build on ci.jenkins.io directly. If you want to rebuild a PR, you can close the PR, wait a few seconds, and then reopen it. Assuming ci.jenkins.io is up and running, that will trigger a new build.

PR #275, opened 2019-11-01
Fix possible typo

ketan:
`Flushable` is in `java.io` package, and not `java.lang`


ketan:
Closing. It seems #268 addresses this.

PR #276, opened 2019-11-08
[JENKINS-39742] Make SecureGroovyScript and ClasspathEntry serializable

FrankS77:
Fix [JENKINS-39742](https://issues.jenkins-ci.org/browse/JENKINS-28732).
Active Choice Plugin should honor ParameterDefinition serializability (was: Active Choice Plugin in Pipelines throw NotSerializableException).
It adds "implements Serializable" to the Java classes SecureGroovyScript and ClasspathEntry.
We can so support better usages in Jenkins pipelines.
I updated Jenkins to 2.164.3 because there were some compilation problems with Jenkins version 2.60
All maven tests were successful and I tested it on my Jenkins. Please have a look.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
   <properties>
     <revision>1.67</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.60.3</jenkins.version>
+    <jenkins.version>2.180</jenkins.version>
     <java.level>8</java.level>
     <configuration-as-code.version>1.30</configuration-as-code.version>
   </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
   <properties>
     <revision>1.67</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.164.3</jenkins.version>
+    <jenkins.version>2.60.3</jenkins.version>
     <java.level>8</java.level>
     <configuration-as-code.version>1.30</configuration-as-code.version>
   </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
   <properties>
     <revision>1.67</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.180</jenkins.version>
+    <jenkins.version>2.164.3</jenkins.version>
     <java.level>8</java.level>
     <configuration-as-code.version>1.30</configuration-as-code.version>
   </properties>
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
@@ -145,7 +145,7 @@ public final class SecureGroovyScript extends AbstractDescribableImpl<SecureGroo
             return;
         }
         if (LOGGER.isLoggable(Level.FINER)) {
-          LOGGER.log(Level.FINER, "found {0}", String.valueOf(loader));
+            LOGGER.log(Level.FINER, "found {0}", String.valueOf(loader));
         }
         if (loader instanceof GroovyClassLoader) {
             GroovyClassLoader gcl = (GroovyClassLoader) loader;
@@ -225,7 +225,7 @@ public final class SecureGroovyScript extends AbstractDescribableImpl<SecureGroo
             if (encounteredLoader != loader) {
                 it.remove();
                 if (LOGGER.isLoggable(Level.FINEST)) {
-                  LOGGER.log(Level.FINEST, "ignoring {0} with loader {1}", new Object[] {klazz, /* do not hold from LogRecord */String.valueOf(encounteredLoader)});
+                    LOGGER.log(Level.FINEST, "ignoring {0} with loader {1}", new Object[] {klazz, /* do not hold from LogRecord */String.valueOf(encounteredLoader)});
                 }
             }
         }
@@ -288,7 +288,7 @@ public final class SecureGroovyScript extends AbstractDescribableImpl<SecureGroo
                 if (iterator.next().getKey().get() == clazz) {
                     iterator.remove();
                     if (LOGGER.isLoggable(Level.FINER)) {
-                      LOGGER.log(Level.FINER, "cleaning up {0} from ObjectStreamClass.Caches.{1}", new Object[] {clazz.getName(), cacheFName});
+                        LOGGER.log(Level.FINER, "cleaning up {0} from ObjectStreamClass.Caches.{1}", new Object[] {clazz.getName(), cacheFName});
                     }
                     break;
                 }
@@ -323,12 +323,12 @@ public final class SecureGroovyScript extends AbstractDescribableImpl<SecureGroo
         List<ClasspathEntry> cp = getClasspath();
         if (!cp.isEmpty()) {
             List<URL> urlList = new ArrayList<URL>(cp.size());
-            
+
             for (ClasspathEntry entry : cp) {
                 ScriptApproval.get().using(entry);
                 urlList.add(entry.getURL());
             }
-            
+
             loader = urlcl = new ClasspathURLClassLoader(urlList.toArray(new URL[urlList.size()]), loader);
         }
         boolean canDoCleanup = false;
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
@@ -35,37 +35,12 @@ import hudson.model.Descriptor;
 import hudson.model.Item;
 import hudson.model.TaskListener;
 import hudson.util.FormValidation;
-
-import java.beans.Introspector;
-import java.lang.ref.Reference;
-import java.lang.ref.WeakReference;
-import java.lang.reflect.Field;
-import java.lang.reflect.Method;
-import java.net.URL;
-import java.net.URLClassLoader;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.ConcurrentMap;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-import javax.annotation.CheckForNull;
-import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import org.codehaus.groovy.control.CompilationUnit;
 import org.codehaus.groovy.control.CompilerConfiguration;
 import org.codehaus.groovy.control.SourceUnit;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException;
-import org.jenkinsci.plugins.scriptsecurity.scripts.ApprovalContext;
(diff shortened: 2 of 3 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -466,4 +456,14 @@ public FormValidation doCheckScript(@QueryParameter String value, @QueryParamete
 
     }
 
-}
+    @SuppressWarnings("unchecked")
+    public Descriptor<SecureGroovyScript> getDescriptor() {
```
dwnusbaum: This is already implemented by the superclass, so I think you can delete it, right?

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -35,52 +35,42 @@
...
 import org.codehaus.groovy.control.CompilationUnit;
 import org.codehaus.groovy.control.CompilerConfiguration;
 import org.codehaus.groovy.control.SourceUnit;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException;
-import org.jenkinsci.plugins.scriptsecurity.scripts.ApprovalContext;
-import org.jenkinsci.plugins.scriptsecurity.scripts.ClasspathEntry;
-import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
-import org.jenkinsci.plugins.scriptsecurity.scripts.UnapprovedClasspathException;
-import org.jenkinsci.plugins.scriptsecurity.scripts.UnapprovedUsageException;
+import org.jenkinsci.plugins.scriptsecurity.scripts.*;
```
dwnusbaum: Please don't reorganize imports unless your PR actually needs to change all of the import lines or is a formatting-changes-only PR, and please don't use wildcard imports in non-test code.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -155,7 +145,7 @@ private static void cleanUpLoader(ClassLoader loader, Set<ClassLoader> encounter
             return;
         }
         if (LOGGER.isLoggable(Level.FINER)) {
-          LOGGER.log(Level.FINER, "found {0}", String.valueOf(loader));
+            LOGGER.log(Level.FINER, "found {0}", String.valueOf(loader));
```
dwnusbaum: Please don't modify whitespace on lines you otherwise didn't change.

pom.xml
```diff
@@ -16,7 +16,7 @@
   <properties>
     <revision>1.67</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.60.3</jenkins.version>
+    <jenkins.version>2.164.3</jenkins.version>
```
dwnusbaum: I don't think the changes in this PR require this bump. What compilation errors were you getting previously?

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
FrankS77: First of all, thanks for checking my PR. I am using Jenkins and the Active Choice Plugin for years but Jenkins Pipeline only for one week. I got a NotSerializableException as I tried this:

`
import org.biouno.unochoice.model.GroovyScript
import org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SecureGroovyScript
import org.biouno.unochoice.CascadeChoiceParameter


node {  
     
    stage('SetParams') { 
        try {
    def sec = new GroovyScript(
        new SecureGroovyScript(
            """
             if ("first".equals(NormalChoice)) {
                 return ['yesFirst1', 'yesFirst2' ] 
             } else {
                 return ['notFirst1', 'notFirst2']
             }
            """, 
            true, 
            []
        ), 
        new SecureGroovyScript(
            "return ['Error']", 
            true, 
            []
        )  
    )
     def newParams = []
     newParams += [$class: 'ChoiceParameterDefinition', name: 'NormalChoice', choices: ['first', 'second']]
     newParams += new CascadeChoiceParameter('Param2', 'description', 'choice-parameter-5', sec, 'PT_SINGLE_SELECT', 'NormalChoice', false, 1)

     props = []
     props += [$class: 'ParametersDefinitionProperty', parameterDefinitions: newParams]
     properties(properties: props)
    } catch (Exception e) {
        println("Exception: ${e}")
    }
 
    }
    stage('Build') { 
       echo 'Building..'
    }
    stage('Deploy') { 
        echo 'Deploying....'
    }
}`

With my changes, I got rid of that exception. I will analyze all your comments and come back to you with questions (I guess).
Thanks!

FrankS77: Hey,
I changed all the things you mentioned but after setting jenkins.version back to 2.60.3
I got a NoSuchFileException "hudson.Extension" in the target folder (with Intellij) and many test failures like:

> Caused by: com.gargoylesoftware.htmlunit.FailingHttpStatusCodeException: 500 Server Error for http://localhost:56777/jenkins/static/e6ca9e18/assets/jquery-detached/jsmodules/jquery2.js
        at com.gargoylesoftware.htmlunit.WebClient.throwFailingHttpStatusCodeExceptionIfNecessary(WebClient.java:590)
        at com.gargoylesoftware.htmlunit.html.HtmlPage.loadJavaScriptFromUrl(HtmlPage.java:1034)
        at com.gargoylesoftware.htmlunit.html.HtmlPage.loadExternalJavaScriptFile(HtmlPage.java:975)
        at com.gargoylesoftware.htmlunit.html.HtmlScript.executeScriptIfNeeded(HtmlScript.java:371)
        at com.gargoylesoftware.htmlunit.html.HtmlScript$2.execute(HtmlScript.java:246)
        at com.gargoylesoftware.htmlunit.javascript.JavaScriptEngine.doProcessPostponedActions(JavaScriptEngine.java:969)
        ... 54 more

> 
Do you know the cause?

dwnusbaum: In what context do you get the exception, just from building the plugin? Do you get this when you run `mvn clean install -DskipTests` from the command line? Or is this when you run the plugin using `mvn hpi:run` or something?

dwnusbaum: That said, your example usage seems strange to me. I think the idea I mentioned earlier would make more sense:

> Could [CascadeChoiceParameter] instead store the textual script as a String, [store the GroovyScript as a transient field], and then recreate the GroovyScript when deserialized?

FrankS77: Hey,
with mvn hpi:run and mvn clean install -DskipTests I get these hudson.Extension file missing exception and when I manually add this file (got it when I build all with higher jenkins version) then I get (with  mvn hpi:run):

```
 WARNING: Failed to instantiate Key[type=org.jenkinsci.main.modules.sshd.SSHD, annotation=[none]]; skipping this component
com.google.inject.ProvisionException: Unable to provision, see the following errors:

1) Error injecting constructor, java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter
  at org.jenkinsci.main.modules.instance_identity.InstanceIdentity.<init>(InstanceIdentity.java:40)
  while locating org.jenkinsci.main.modules.instance_identity.InstanceIdentity
    for field at org.jenkinsci.main.modules.sshd.SSHD.identity(SSHD.java:35)

```
But maybe I don't need this fix anymore. But I think it is only like a workaround and
I can only imagine why the same pipeline code works (without the fix) when I write it
in another syntax:

```
def sec = [$class: 'GroovyScript',
        script: [$class: 'SecureGroovyScript', script: """ return ['yesFirst1', 'yesFirst2' ]  """, sandbox: true, classpath: null],
        fallbackScript: [$class: 'SecureGroovyScript', script: """ return ['ERROR'] """, sandbox: true, classpath: null]]
```
So this code works without my fix but (see above) code with "def sec = new GroovyScript ..." does not work (get NotSerializableException).

My question now is whether we should continue here. I think it would be nice (also for other developers) to continue. What do you think?
And your suggestion to save the script only as a string would of course be sufficient, but that would be a change in the Active Choice Jenkins plugin.

dwnusbaum: > ```
> def sec = [$class: 'GroovyScript',
>         script: [$class: 'SecureGroovyScript', script: """ return ['yesFirst1', 'yesFirst2' ]  """, sandbox: true, classpath: null],
>         fallbackScript: [$class: 'SecureGroovyScript', script: """ return ['ERROR'] """, sandbox: true, classpath: null]]
> ```

That is the normal databinding syntax, although I'm surprised that it works without the fix. You could clean it up a bit using `@Symbol` on some of the types, but it's going to be awkward either way.

> My question now is whether we should continue here. I think it would be nice (also for other developers) to continue. What do you think?
> And your suggestion to save the script only as a string would of course be sufficient, but that would be a change in the Active Choice Jenkins plugin.

I think that it would be _best_ to change Active Choices Plugin so that it can be used from Pipeline more naturally, but the change here is harmless so I am fine with it. If you update the PR to just add `implements Serializable`, and remove the Jenkins update and formatting changes, we can see if the PR build passes and go from there.

FrankS77:
Hey, I pushed to my branch. Do I have to create another pull request? Sorry this is one of my first pull requests.

dwnusbaum:
@FrankS77 no, this one is fine, we can squash-merge when it's ready. Thanks for making the changes!

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SecureGroovyScript.java
```diff
@@ -466,4 +467,4 @@ public FormValidation doCheckScript(@QueryParameter String value, @QueryParamete
 
     }
 
-}
+}
```
dwnusbaum:
```suggestion
}

```

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ClasspathEntry.java
```diff
@@ -38,6 +38,8 @@
 import hudson.model.Descriptor;
 import hudson.model.Items;
 import hudson.util.FormValidation;
+
```
dwnusbaum:
```suggestion
```

dwnusbaum:
Can you please accept my suggestions via the GitHub UI (or commit them manually) and then I'll squash-merge the PR? I thought I'd be able to do that myself, but I think you unchecked the PR setting that lets me submit changes to your branch.

dwnusbaum:
https://github.com/jenkinsci/script-security-plugin/pull/292 was filed without the whitespace changes, and I just merged that, so I am going to close this PR, but thanks for working on it!

philnalwalker:
Is there a release with this change in it? If not, would you please cut a release? Until there is a release pipeline developers can not utilize the Active Choices plugin CascadeChoiceParameter script functionality due to the missing Serializable implementation on SecureGroovyScript. Thank you!

PR #278, opened 2019-11-14
Whitelist ChangeLogSet.Entry.getComment()

rufoa:
getComment() method was introduced in JENKINS-49202 but was not whitelisted at the time

This PR brings it into alignment with getMsg, getAuthor, etc


rufoa:
Hang on, it looks like JENKINS-49202 was never actually merged

PR #279, opened 2019-11-15
Diagnostics & robustness for corrupt Whitelist.all

jglick:
See https://github.com/jenkinsci/kubernetes-plugin/pull/649 and https://github.com/jenkinsci/workflow-cps-plugin/pull/338. It seems that if a test suite containing several test cases (in one JVM, the default mode) starts a Pipeline build in one case and then immediately exits, the CPS VM thread will continue to run in the background, and when it hits the first sandbox operation it will try to initialize the global `Whitelist`. At that point `Jenkins.instance` may correspond to a different test, just starting up. The extension list will still be empty, so `Whitelist.all()` will deny everything; this will break the build started in the first test, which is irrelevant, but will also _cache_ the bad proxy whitelist for the new test, causing _its_ build to fail with a cryptic error

```
Scripts not permitted to use staticMethod org.jenkinsci.plugins.workflow.cps.Safepoint safepoint. Administrators can decide whether to approve or reject this signature.
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
@@ -32,6 +32,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.Map;
 import java.util.WeakHashMap;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
@@ -77,16 +78,17 @@ public abstract class Whitelist implements ExtensionPoint {
     public static synchronized @Nonnull Whitelist all() {
         Jenkins j = Jenkins.getInstanceOrNull();
         if (j == null) {
-            LOGGER.warning("No Jenkins.instance");
+            LOGGER.log(Level.WARNING, "No Jenkins.instance", new Throwable("here"));
             return new ProxyWhitelist();
         }
         Whitelist all = allByJenkins.get(j);
         if (all == null) {
             ExtensionList<Whitelist> allWhitelists = j.getExtensionList(Whitelist.class);
             if (allWhitelists.isEmpty()) {
-                LOGGER.warning("Empty list of Whitelist");
+                LOGGER.log(Level.WARNING, "No Whitelist instances registered", new Throwable("here"));
+                return new ProxyWhitelist();
             } else {
-                /*TODO LOGGER.fine*/System.err.println("Loading whitelists: " + allWhitelists);
+                LOGGER.fine(() -> "Loading whitelists: " + allWhitelists);
             }
             all = new ProxyWhitelist(allWhitelists);
             allByJenkins.put(j, all);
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/Whitelist.java
```diff
@@ -73,11 +77,18 @@
...
         if (j == null) {
+            LOGGER.warning("No Jenkins.instance");
             return new ProxyWhitelist();
         }
         Whitelist all = allByJenkins.get(j);
         if (all == null) {
-            all = new ProxyWhitelist(j.getExtensionList(Whitelist.class));
+            ExtensionList<Whitelist> allWhitelists = j.getExtensionList(Whitelist.class);
+            if (allWhitelists.isEmpty()) {
+                LOGGER.warning("Empty list of Whitelist");
```
dwnusbaum: A stack trace might be interesting here:

```suggestion
                LOGGER.log(java.util.logging.Level.WARNING, null, new Exception("Empty list of Whitelist"));
```

jglick:
```
java.lang.UnsatisfiedLinkError: C:\openjdk-11\bin\awt.dll: Can't find dependent libraries
	at java.base/java.lang.ClassLoader$NativeLibrary.load0(Native Method)
	at java.base/java.lang.ClassLoader$NativeLibrary.load(ClassLoader.java:2430)
	at java.base/java.lang.ClassLoader$NativeLibrary.loadLibrary(ClassLoader.java:2487)
	at java.base/java.lang.ClassLoader.loadLibrary0(ClassLoader.java:2684)
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2638)
	at java.base/java.lang.Runtime.loadLibrary0(Runtime.java:829)
	at java.base/java.lang.System.loadLibrary(System.java:1867)
	at java.desktop/java.awt.Toolkit$3.run(Toolkit.java:1395)
	at java.desktop/java.awt.Toolkit$3.run(Toolkit.java:1393)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.desktop/java.awt.Toolkit.loadLibraries(Toolkit.java:1392)
	at java.desktop/java.awt.Toolkit.<clinit>(Toolkit.java:1425)
	at java.desktop/java.awt.Font.<clinit>(Font.java:260)
	at hudson.model.MultiStageTimeSeries.<clinit>(MultiStageTimeSeries.java:98)
	at hudson.model.LoadStatistics.<init>(LoadStatistics.java:139)
	at hudson.model.OverallLoadStatistics.<init>(OverallLoadStatistics.java:55)
	at jenkins.model.Jenkins.<init>(Jenkins.java:662)
	at hudson.model.Hudson.<init>(Hudson.java:85)
	at org.jvnet.hudson.test.JenkinsRule.newHudson(JenkinsRule.java:675)
	at …
```

urgh

jglick:
> Not really possible to write a test for this I imagine.

Umm…not easily. You would have to have a new variant of `RestartableJenkinsRule` and then run something after `Jenkins.instance` is set but before Jenkins initialization completes, which is tricky. I think the test would wind up being quite complex, compared to the rather straightforward patch.

PR #282, opened 2020-01-07
Improve scriptApproval ux

jsoref:
The main items are:

* Wrap things in `<p>` so that they don't flow to the right of a label (`:`).
* Offer `None` text of sorts instead of nothing...
* Add extra headings and anchors


jeffret-b:
@jsoref, I've been independently looking into some improvements for this page. It's still in the discussion stage, trying to figure out what makes sense. You've got some good ideas here. I'm working on pulling together some proposals and seeing where we can make some gains with limited efforts. I'll share what I come up with as soon as I can.

Wadeck:
FYI the effort discussed by @jeffret-b led to #300

DuMaM:
I think those changes was approved or they where fixed with recent ui changes
image

PR #290, opened 2020-03-18
Clear static field signatures in ProxyWhitelist.reset

dwnusbaum:
While investigating something else, I noticed that `ProxyWhitelist.reset` didn't clear out static field  signatures when clearing out other types of signatures.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelistTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelistTest.java
@@ -47,6 +47,14 @@ public class ProxyWhitelistTest {
         assertFalse(pw2.permitsMethod(Object.class.getMethod("hashCode"), "x", new Object[0]));
     }
 
+    @Test public void resetStaticField() throws Exception {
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelistTest.java
```diff
@@ -47,6 +47,14 @@
         assertFalse(pw2.permitsMethod(Object.class.getMethod("hashCode"), "x", new Object[0]));
     }
 
+    @Test public void resetStaticField() throws Exception {
+        ProxyWhitelist pw1 = new ProxyWhitelist(new StaticWhitelist(new StringReader("staticField java.util.Collections EMPTY_LIST")));
+        ProxyWhitelist pw2 = new ProxyWhitelist(pw1);
+        assertTrue(pw2.permitsStaticFieldGet(Collections.class.getField("EMPTY_LIST")));
+        pw1.reset(Collections.<Whitelist>emptySet());
+        assertFalse(pw2.permitsStaticFieldGet(Collections.class.getField("EMPTY_LIST")));
```
dwnusbaum: Fails without the patch, because the old entry is not removed.

PR #292, opened 2020-05-04
implementing serializable in SecureGroovyScript & ClasspathEntry

aleiphoenix:
Same as #276 . It's been 5 months since last reply. 

Only add `Serializable`. This will make plugins like active-choices-plugin compatible to pipeline script.


dwnusbaum:
Rebuilding now that #293 was merged.

PR #298, opened 2020-05-11
[JENKINS-42214] SandboxInterceptor must account for static members being accessed via objects instead of class references

dwnusbaum:
See [JENKINS-42214](https://issues.jenkins-ci.org/browse/JENKINS-42214). Fixing the issue going forward is easy enough, but I need to check the upgrade behavior for existing field signatures in their `scriptApproval.xml` that would be considered invalid after this change.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
@@ -44,6 +44,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.lang.reflect.Modifier;
 import org.codehaus.groovy.runtime.DateGroovyMethods;
 import org.codehaus.groovy.runtime.DefaultGroovyMethods;
 import org.codehaus.groovy.runtime.EncodingGroovyMethods;
@@ -157,12 +158,12 @@ final class SandboxInterceptor extends GroovyInterceptor {
             throw new MissingMethodException(method, receiver.getClass(), args);
         } else if (StaticWhitelist.isPermanentlyBlacklistedMethod(m)) {
             throw StaticWhitelist.rejectMethod(m);
-        } else if (whitelist.permitsMethod(m, receiver, args)) {
+        } else if (permitsMethod(whitelist, m, receiver, args)) {
             return super.onMethodCall(invoker, receiver, method, args);
         } else if (method.equals("invokeMethod") && args.length == 2 && args[0] instanceof String && args[1] instanceof Object[]) {
             throw StaticWhitelist.rejectMethod(m, EnumeratingWhitelist.getName(receiver.getClass()) + " " + args[0] + printArgumentTypes((Object[]) args[1]));
         } else {
-            throw StaticWhitelist.rejectMethod(m);
+            throw rejectMethod(m);
         }
     }
 
@@ -206,11 +207,11 @@ final class SandboxInterceptor extends GroovyInterceptor {
                 ? setterMethods.get(0) // If there is only a single setter, the argument will be cast to match the declared parameter type.
                 : GroovyCallSiteSelector.method(receiver, setter, valueArg); // If there are multiple setters, MultipleSetterProperty just calls invokeMethod.
         if (setterMethod != null) {
-            if (whitelist.permitsMethod(setterMethod, receiver, valueArg)) {
+            if (permitsMethod(whitelist, setterMethod, receiver, valueArg)) {
                 preCheckArgumentCasts(setterMethod, valueArg);
                 return super.onSetProperty(invoker, receiver, property, value);
             } else if (rejector == null) {
-                rejector = () -> StaticWhitelist.rejectMethod(setterMethod);
+                rejector = () -> rejectMethod(setterMethod);
             }
         }
         Object[] propertyValueArgs = new Object[] {property, value};
@@ -223,13 +224,13 @@ final class SandboxInterceptor extends GroovyInterceptor {
                 rejector = () -> StaticWhitelist.rejectMethod(setPropertyMethod, receiver.getClass().getName() + "." + property);
             }
         }
-        final Field instanceField = GroovyCallSiteSelector.field(receiver, property);
-        if (instanceField != null) {
-            if (whitelist.permitsFieldSet(instanceField, receiver, value)) {
-                Checker.preCheckedCast(instanceField.getType(), value, false, false, false);
+        final Field field = GroovyCallSiteSelector.field(receiver, property);
+        if (field != null) {
+            if (permitsFieldSet(whitelist, field, receiver, value)) {
+                Checker.preCheckedCast(field.getType(), value, false, false, false);
                 return super.onSetProperty(invoker, receiver, property, value);
             } else if (rejector == null) {
-                rejector = () -> StaticWhitelist.rejectField(instanceField);
+                rejector = () -> rejectField(field);
             }
         }
         if (receiver instanceof Class) {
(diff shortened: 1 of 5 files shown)


jglick (approved):

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -546,4 +547,39 @@ private static String printArgumentTypes(Object[] args) {
         }
     }
 
+    private static boolean permitsFieldGet(@NonNull Whitelist whitelist, @NonNull Field field, @NonNull Object receiver) {
+        if (Modifier.isStatic(field.getModifiers())) {
+            return whitelist.permitsStaticFieldGet(field);
+        }
+        return whitelist.permitsFieldGet(field, receiver);
+    }
```
dwnusbaum, line 555: We could put this methods in `Whitelist` itself, but then we would need to come up with a different name.

In retrospect I think we should not have bothered with an API-level distinction between static and instance fields and methods in `Whitelist` and just made implementations check modifiers themselves.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -546,4 +547,39 @@ private static String printArgumentTypes(Object[] args) {
...
+        }
+        return whitelist.permitsMethod(method, receiver, args);
+    }
+
+    public static RejectedAccessException rejectMethod(@NonNull Method m) {
+        if (Modifier.isStatic(m.getModifiers())) {
+            return StaticWhitelist.rejectStaticMethod(m);
+        }
+        return StaticWhitelist.rejectMethod(m);
+    }
```
dwnusbaum, line 576: Similarly, we could put this in `StaticWhitelist`, but we would need to change the name. Another option could be to change the behavior of the current `rejectMethod` and deprecate `rejectStaticMethod`, but we would need to take a closer look at code like https://github.com/jenkinsci/permissive-script-security-plugin/blob/45af9a18a739fbb275093da5effcf1d2b7637a05/src/main/java/org/jenkinsci/plugins/permissivescriptsecurity/PermissiveWhitelist.java#L112 to understand whether that change could cause issues.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -911,6 +912,8 @@ public static void assertRejected(Whitelist whitelist, String expectedSignature,
             // GroovyRuntimeException.
             if (x.getCause() instanceof RejectedAccessException) {
                 errors.checkThat(x.getMessage(), ((RejectedAccessException)x.getCause()).getSignature(), is(expectedSignature));
+            } else {
+                errors.addError(x);
```
dwnusbaum, line 916: This corrects a mistake in 85d16bd851b355458f6396dbc4f92c3e0eebcd72. Thankfully it did not mask any existing issues.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -1650,6 +1653,53 @@ public void booleanCasts() throws Throwable {
         assertEvaluate(new GenericWhitelist(), true, "('a' =~ '.*') as Boolean");
     }
 
+    @Test
+    public void accessStaticMembersViaInstance() throws Throwable {
+        String fqcn = HasStaticMembers.class.getName();
+        // Make sure that we report the correct signature in RejectedAccessException.
```
dwnusbaum, line 1659: All of these cases previously threw instances of `RejectedAccessException` with bogus non-static signatures.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -1650,6 +1653,53 @@ public void booleanCasts() throws Throwable {
...
+        assertRejected(new AnnotatedWhitelist(), "staticMethod " + fqcn + " isBAZ", "def o = new " + fqcn + "(); o.BAZ");
+        assertRejected(new AnnotatedWhitelist(), "staticField " + fqcn + " FOO", "def o = new " + fqcn + "(); o.FOO = 1");
+        assertRejected(new AnnotatedWhitelist(), "staticMethod " + fqcn + " setBAR int", "def o = new " + fqcn + "(); o.BAR = 2");
+        assertRejected(new AnnotatedWhitelist(), "staticField " + fqcn + " FOO", "def o = new " + fqcn + "(); o.@FOO");
+        assertRejected(new AnnotatedWhitelist(), "staticField " + fqcn + " FOO", "def o = new " + fqcn + "(); o.@FOO = 1");
+        assertRejected(new AnnotatedWhitelist(), "staticMethod " + fqcn + " method", "def o = new " + fqcn + "(); o.method()");
+        assertRejected(new AnnotatedWhitelist(), "staticMethod " + fqcn + " getAt int", "def o = new " + fqcn + "(); o[0]");
+        assertRejected(new AnnotatedWhitelist(), "staticMethod " + fqcn + " putAt int java.lang.Object", "def o = new " + fqcn + "(); o[0] = null");
+        // Make sure that we check for the correct signature in allowlists.
```
dwnusbaum, line 1670: All of these cases previously failed because bogus non-static signatures were used when checking allowlists.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -1650,6 +1653,53 @@ public void booleanCasts() throws Throwable {
         assertEvaluate(new GenericWhitelist(), true, "('a' =~ '.*') as Boolean");
     }
 
+    @Test
```
dwnusbaum, line 1656:
```suggestion
    @Issue("JENKINS-42214")
    @Test
```

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -414,11 +415,11 @@ private interface Rejector {
         Rejector rejector = null;
         Field field = GroovyCallSiteSelector.field(receiver, attribute);
         if (field != null) {
-            if (whitelist.permitsFieldSet(field, receiver, value)) {
+            if (permitsFieldSet(whitelist, field, receiver, value)) {
                 Checker.preCheckedCast(field.getType(), value, false, false, false);
                 return super.onSetAttribute(invoker, receiver, attribute, value);
             } else {
-                throw StaticWhitelist.rejectField(field);
+                rejector = () -> rejectField(field);
```
jglick, line 422: Intentionally changing this to delayed?

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
jglick: I guess so. A bit tricky to change now though, since you would need to use `Util.isOverridden` to accommodate the fact that there are both implementors and (apparently) callers of `Whitelist` in other plugins.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
dwnusbaum: Ah yeah, I forgot to note that this was also an oversight in https://github.com/jenkinsci/script-security-plugin/commit/85d16bd851b355458f6396dbc4f92c3e0eebcd72. I think the effect is that the attempt to make static fields accessible via attribute syntax (`.@`) in that commit did not work correctly when modifying the attribute, only when getting it. Before that commit it did not work in either case, and after this commit it will work in both cases.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
dwnusbaum: Yeah, I do not really think that it is worth making API changes at this point.

dwnusbaum:
Rebuilding.

Also, I checked the behavior for cases where users have an invalid approved signature in their `scriptApproval.xml`. Previously, the behavior was wrong, but it was wrong both when checking the signature against allowlists and when reporting the rejection, so things worked fine as long as you had assertions disabled. If you approved some invalid signature before this update, then after this update you will need to approve the correct signature.

We could check `Signature.exists` during deserialization and try to migrate existing signatures that hit this case, but I'm not sure that the complexity is worth it for what is probably an uncommon case. I think it should be good enough to mention that you may need to approve some signatures after the update in the release notes.

jglick:
> Rebuilding

BTW **Re-run** from **Checks** should work.

jglick:
> I think it should be good enough to mention that you may need to approve some signatures after the update in the release notes.

Agreed

dwnusbaum:
> BTW Re-run from Checks should work.

Hmm, it never seems to work for me. I will try that next time though.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
dwnusbaum: Actually I think the problems were more limited. If you enter this method with code like `MyClass.staticField = x`, then `GroovyCallSiteSelector.field` could only find fields declared in `java.lang.Class` itself or its supertypes, so `field != null` could only be true if your static field has the same name as a field declared in `java.lang.Class`. In other cases, the code will fall through to the new case from https://github.com/jenkinsci/script-security-plugin/commit/85d16bd851b355458f6396dbc4f92c3e0eebcd72 (which has test coverage), and will work correctly. So for example, if you have a static field named `name`, `MyClass.@name = x` will fail because `field java.lang.Class name` is not allowed to be used. It didn't work correctly before https://github.com/jenkinsci/script-security-plugin/commit/85d16bd851b355458f6396dbc4f92c3e0eebcd72 either though, so impact is minimal.

The other difference would be that if you had code like `object.staticField` for some field that you had approved as `staticField Foo foo` previously, the old code would check if `field Foo foo` was approved, and if not, it would reject it as `field ...`, rather than checking if `staticField ...` is allowed. That also would have been broken before https://github.com/jenkinsci/script-security-plugin/commit/85d16bd851b355458f6396dbc4f92c3e0eebcd72 though.

PR #299, opened 2020-05-11
Add result color related methods to Jenkins whitelist

basil:
Adds three result color related methods to the Jenkins whitelist. I took a quick look at these methods and they seemed safe to me; however, I do not have a full understanding of how to evaluate safety, so if these methods are unsafe I would appreciate an explanation as to why.

For background, I use these methods in my pipelines as follows:

```groovy
/**
 * Return a Slack color from a given Jenkins result.
 *
 * @param resultStr The Jenkins job result, as returned by {@link RunWrapper#getResult()}. Typically
 * this is the value of {@code currentBuild.result}.
 * @return A valid Slack color. This is typically one of the predefined Slack colors or otherwise
 * the value of {@link BallColor#getHtmlBaseColor()}.
 */
def getSlackColor(String resultStr) {
  Result result = Result.fromString(resultStr)
  if (result == Result.SUCCESS) {
    return 'good'
  } else if (result == Result.UNSTABLE) {
    return 'warning'
  } else if (result == Result.FAILURE) {
    return 'danger'
  } else {
    return result.color.htmlBaseColor
  }
}
```

And I call the `getSlackColor` method as follows:

```
slackSend(channel: '#foo', color: getSlackColor(currentBuild.result), message: '[…]')
```

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/jenkins-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/jenkins-whitelist
@@ -1,6 +1,7 @@
 # Useful Jenkins API methods:
 new hudson.EnvVars
 method hudson.model.AbstractBuild getEnvironments
+method hudson.model.BallColor getHtmlBaseColor
 staticMethod hudson.model.Environment create hudson.EnvVars
 method hudson.model.Item getParent
 method hudson.model.ModelObject getDisplayName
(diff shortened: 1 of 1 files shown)


dwnusbaum:
Looks like there was an infra issue, rebuilding:

> [2020-05-11T22:57:58.849Z] OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00000000fed00000, 19922944, 0) failed; error='Cannot allocate memory' (errno=12)
> [2020-05-11T22:57:58.849Z] # There is insufficient memory for the Java Runtime Environment to continue.
> [2020-05-11T22:57:58.849Z] # Native memory allocation (mmap) failed to map 19922944 bytes for committing reserved memory.

PR #300, opened 2020-05-25
[JENKINS-62448] Enhance information displayed in approval page

Wadeck:
With this PR, we are storing more information than before. The change is incremental, it means that if you have some hashes already approved, they will just not have any metadata until used. When used, the metadata will be saved.

It means, after some weeks / months, you will know easily which entries are approved but not used (or no longer used).

If you have troubles with this PR, you can disable this feature by setting the System Property: `org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval.metadataGathering` to false.

_There is no test addition at the moment, I want to have feedbacks on the approach before investing time on the testing part._

## Testing notes
Having pipeline installed will ease the testing.
Also, if you have an instance with already some approved items, it could be useful. I did some manual testing of "migration" without any issues but not 100% sure to have covered all corner cases.

### To generate a full script approval request
You can create a new pipeline with a user without RUN_SCRIPT permission (just Job/Configure). Configure the pipeline content locally ("Pipeline script"), and do not check the "Use Groovy Sandbox". A simple `echo 'hello from pipeline'` is sufficient.
Then run the pipeline.

### To generate a signature approval request
Same as full script but you check the Groovy Sandbox.
`new File(".")` will trigger the signature for `new java.io.File java.lang.String`.

### To generate a class entry approval request
It's a bit more complicated (or at least I was not able to find a better way). You can install EnvInject plugin. Then inside a Freestyle project, you configure the "Build Environment", with "Inject environment variables to the build process". As Groovy Script, put anything and use "Additional classpath", you can use the regular Artifactory of Jenkins to provide links. 
I was using: https://repo.jenkins-ci.org/javanet2-cache/org/jvnet/hudson/main/maven-plugin/1.301/maven-plugin-1.301.hpi.

<details>

<summary>Screenshots</summary>

### Full script pending approvals

script-security-approval

### Full script already approved

script-security-approved

</details>

### Scope
The objective of the PR is only to provide metadata and new features for the Full Script approval process as it's the most annoying part. The two other categories are let aside (at least for the moment). Their script parts is let unmodified, just split and moved.

### Reviewers

@jsoref as you worked on #282
@josephbrueggen for the UI/UX
@fqueiruga for the JavaScript review
@dwnusbaum as maintainer of the plugin
@jeffret-b as you worked on this effort too


src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.jelly
```diff
@@ -0,0 +1,204 @@
...
+                <j:when test="${hashAndMetadataList.isEmpty()}">
+                    <p class="no-approved-scripts">
+                        <i>${%noApprovedScripts}</i>
+                    </p>
+                </j:when>
+                <j:otherwise>
+                    <table class="pane bigtable">
+                        <tr>
+                            <th width="1%"><!-- checkbox --></th>
+                            <th tooltip="${%approvedHash_tooltip}" class="info-cursor">${%approvedHash}</th>
```
timja, line 25: Why does the user care about the hash?

Wadeck: In the current version only the hash is stored in the XML. By providing it in the table display I initially thought it could be good for the user to know which hash correspond to what scripts. The metadata and the hashes are separated.
But as a second thought with your question, I am no longer certain it's a good idea :p 

Potentially hidden the hash as just a HTML attribute could be sufficient for advance users and keeping the UI simpler... good point :)

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/classPath_pending.jelly
```diff
@@ -0,0 +1,74 @@
...
+        </p>
+        <div id="pendingClasspathEntries">
+            Classpath entries pending approval. (Beware of remote URLs, workspace files, or anything else that might change without your notice.)
+        </div>
+    </div>
+    
+    <script>
+        var mgr = <st:bind value="${it}"/>;
+        function renderPendingClasspathEntries(pendingClasspathEntries) {
+            if (pendingClasspathEntries.length == 0) {
```
fqueiruga, line 15: Is this new inline JS and prototype code? I would suggest extracting this and use vanilla JS

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,63 @@
...
+noUsageCountSinceMetadata=N/A
+noUsageCountSinceMetadata_tooltip=It was not used since the introduction of metadata.
+notUsedSinceMetadata=N/A
+notUsedSinceMetadata_tooltip=It was not used since the introduction of metadata.
+notApprovedSinceMetadata=N/A
+notApprovedSinceMetadata_tooltip=It was not approved since the introduction of metadata.
+noKnownApproverSinceMetadata=N/A
+noKnownApproverSinceMetadata_tooltip=It was not approved since the introduction of metadata.
+
+revokeApproval=Revoke all selected
```
fqueiruga, line 52: I would prefer that the buttons read "Revoke/Approve selected", skipping the "all". Thoughts of a native speaker @timja ?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.jelly
timja: Yeah not sure, the only reason I can see it for a user as an ID.
I would lean towards less is more though

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
timja: Without seeing the full context I would lean towards one of:

* Revoke
* Remove
* Remove selected

Revoke isn't really a call to action text you would normally see in a UI

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/classPath_approved.jelly
```diff
@@ -0,0 +1,75 @@
...
+    <div class="pane-frame">
+        <p>Classpath entries already approved:</p>
+        <p id="approvedClasspathEntries-none">
+            No approved classpath entries.
+        </p>
+        <div id="approvedClasspathEntries">
+        </div>
+        <p id="approvedClasspathEntries-clear">
+            You can also remove all previous classpath entry approvals:
+            <button onclick="if (confirm('Really delete all approvals? Any existing scripts using a classpath will need to be rerun and entries reapproved.')) {clearApprovedClasspathEntries()}">Clear Classpath Entries</button>
```
jsoref, line 12:
```suggestion
            You can also remove approval for all previous classpath entries:
            <button onclick="if (confirm('Really clear all approvals? Any existing scripts using a classpath will need to be rerun and their entries reapproved.')) {clearApprovedClasspathEntries()}">Clear Classpath Entries</button>
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/classPath_approved.jelly
```diff
@@ -0,0 +1,75 @@
...
+                        <button class="delete" onclick="denyApprovedClasspathEntry('${pcp.hash}')">Delete</button> 
+                        ${acp.hash} (${acp.path})
+                    </p>
+                 */
+                approvedClasspathEntries.each(function(e) {
+                    var block = new Element('p', { 'id': 'acp-' + e.hash });
+                    var deleteButton = new Element('button', { 'class': 'delete', 'hash': e.hash});
+                    deleteButton.insert('Delete');
+                    deleteButton.observe('click', function() {
+                        if (confirm('Really delete this approved classpath entry? Any existing scripts using it will need to be rerun and the entry reapproved.')) {
```
jsoref, line 40:
```suggestion
                        if (confirm('Really remove this approved classpath entry? Any existing scripts using it will need to be rerun and their entry reapproved.')) {
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/classPath_pending.jelly
```diff
@@ -0,0 +1,74 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler">
+    <div class="pane-frame">
+        <p id="pendingClasspathEntries-none">
+            No pending classpath entry approvals.
+        </p>
+        <div id="pendingClasspathEntries">
+            Classpath entries pending approval. (Beware of remote URLs, workspace files, or anything else that might change without your notice.)
```
jsoref, line 8:
```suggestion
            Classpath entries pending approval. (Beware of remote URLs, workspace files, or anything else that might change without you noticing.)
```

I'm not sure if this is what it means....

`change without notice` is a technical thing, and I'm not sure it calls to mind the right concern.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,63 @@
...
+# approved scripts
+metadataGatheringDisabled=The metadata gathering was disabled. \
+  <p>You can re-enable it permanently by removing the system property: <br />\
+  <code>org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval.metadataGathering</code> </p>\
+  <p>or temporarily (until restart) by running this in the Script Console: <br />\
+  <code>org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval.METADATA_GATHERING = true</code></p>
+
+## table headers
+approvedHash=Hash
+approvedHash_tooltip=The hash could be seen as a signature of the script. It's used as an identifier to reference the content of the script.
```
jsoref, line 12: I'm not a fan of calling a hash a signature.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,63 @@
+approvedScripts=Approved scripts
+
+# approved scripts
+metadataGatheringDisabled=The metadata gathering was disabled. \
+  <p>You can re-enable it permanently by removing the system property: <br />\
+  <code>org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval.metadataGathering</code> </p>\
+  <p>or temporarily (until restart) by running this in the Script Console: <br />\
+  <code>org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval.METADATA_GATHERING = true</code></p>
```
jsoref, line 8: I don't understand this.

What disabled it? (was it the preceding user action?) - in which case, I'd probably drop `was`.
Naively, I'd change `The metadata` to `Metadata`.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,63 @@
...
+
+## table headers
+approvedHash=Hash
+approvedHash_tooltip=The hash could be seen as a signature of the script. It's used as an identifier to reference the content of the script.
+approvedExpandCode=Code
+approvedLanguage=Language
+approvedScriptLength=Length
+approvedScriptLength_tooltip=Length of the script in characters
+usageCount=# of uses
+usageCount_tooltip=Usage since the introduction of metadata
```
jsoref, line 18:
```suggestion
usageCount_tooltip=Uses since the introduction of metadata
```

`usage` and `use` are distinct things, if `use` applies, it should be used consistently throughout to apply to a given topic. In the rare cases where it doesn't, you can then use `usage`, but again, consistently for a given concept.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,63 @@
...
+lastTimeUsed=Date of last use
+wasPreapproved_header_tooltip=The presence of a green tick means the script was approved during the configuration. \
+  Otherwise it was created by a user without the permission to run scripts and was then approved using this page.
+lastApprovalTime=Date of the last approval
+lastApprovalTime_tooltip=A single script can be approved multiple times. It occurs every time the related configuration is saved.
+lastKnownApproverLogin=Last approver
+
+## table content
+noRequesterSinceMetadata=N/A
+noRequesterSinceMetadata_tooltip=The requester was not recorded. It could be due to the request being created before the metadata introduction or the code calling this script does not provide the user information.
```
jsoref, line 28:
```suggestion
noRequesterSinceMetadata_tooltip=The requester was not recorded. It could be due to the request being created before the metadata introduction or the code calling this script did not provide user information.
```
Note: this line is really long and hard to suggest against w/ GitHub (I accidentally left this page while trying to scroll to comment)

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,63 @@
...
+noRequestApprovalTimeSinceMetadata=N/A
+noRequestApprovalTimeSinceMetadata_tooltip=It was requested before the introduction of metadata.
+noContextItemSinceMetadata=N/A
+noContextItemSinceMetadata_tooltip=The context item was not provided by the code asking for the approval.
```
jsoref, line 32:
```suggestion
noContextItemSinceMetadata_tooltip=The context item was not provided by the code asking for approval.
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,63 @@
...
+
+emptyMetadata=There is no metadata for this approval
+emptyMetadata_tooltip=If the script is used after the metadata introduction, some metadata will be displayed here. \
+  It means that an unused approval will not have metadata and could be reasonably revoked after a certain period of time.
+noScriptCodeSinceMetadata=N/A
+noScriptCodeSinceMetadata_tooltip=The approval was given while the metadata were not gathered.
```
jsoref, line 38:
```suggestion
noScriptCodeSinceMetadata_tooltip=The approval was given while the metadata was not gathered.
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
jsoref: Pictures wanted

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,63 @@
...
+notUsedSinceMetadata=N/A
+notUsedSinceMetadata_tooltip=It was not used since the introduction of metadata.
+notApprovedSinceMetadata=N/A
+notApprovedSinceMetadata_tooltip=It was not approved since the introduction of metadata.
+noKnownApproverSinceMetadata=N/A
+noKnownApproverSinceMetadata_tooltip=It was not approved since the introduction of metadata.
+
+revokeApproval=Revoke all selected
+revokeApproval_tooltip=Revoking an approved script can have an impact on the jobs using it. The next time an attempt to execute the script is made, a new approval will be required.
+revokeApproval_confirm=Really delete all selected approvals? Any existing scripts will need to be requeued and reapproved.
```
jsoref, line 54:
```suggestion
revokeApproval_confirm=Really remove all selected approvals? Any existing scripts will need to be requeued and reapproved.
```

I'm not a fan of `selected approvals` -- in general, I'd prefer `approval of selected ...`

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,63 @@
...
+approvedNoSelectedItemError=Please select at least one line.
+
+approvedExpand=expand code
+approvedExpand_tooltip=Expand the source code visualization
+approvedMinimize=minimize code
+approvedMinimize_tooltip=Minimize the source code visualization
+scriptContentLoadingMessage=Loading...
+scriptContentLoadError=Impossible to find the script content. It was perhaps removed manually. On next usage this should be self-repaired.
+scriptContentLoadError403=Impossible to request the script content due to lack of authorization. Try reloading the page.
```
jsoref, line 63:
```suggestion
scriptContentLoadError=Unable to find the script content. It was perhaps removed manually. On next use this should be self-repaired.
scriptContentLoadError403=Unable to request the script content due to lack of authorization. Try reloading the page.
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_pending.properties
```diff
@@ -0,0 +1,38 @@
...
+pendingHash=Hash
+pendingHash_tooltip=The hash could be seen as a signature of the script. It's used as an identifier to reference the content of the script.
+pendingExpandCode=Code
+pendingLanguage=Language
+pendingScriptLength=Length
+pendingScriptLength_tooltip=Length of the script in characters
+contextUser=Requester
+contextItem=Context
+requestApprovalTime=Request date
+requestApprovalTime_tooltip=Inform about when the approval request was done.
```
jsoref, line 14:
```suggestion
requestApprovalTime_tooltip=Inform about when the approval request was made.
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_pending.properties
```diff
@@ -0,0 +1,38 @@
...
+
+## table content
+noRequesterSinceMetadata=N/A
+noRequesterSinceMetadata_tooltip=The requester was not recorded. It could be due to the request being created before the metadata introduction or the code calling this script does not provide the user information.
```
jsoref, line 18:
```suggestion
noRequesterSinceMetadata_tooltip=The requester was not recorded. It could be due to the request being created before the metadata introduction or the code calling this script did not provide user information.
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_pending.properties
```diff
@@ -0,0 +1,38 @@
...
+denyPending_tooltip=Denying a pending script will just remove it from the pending list. If there are new script execution attempts, it will reappear in this list.
+denyPending_confirm=Really deny all selected scripts? If they are triggered again, they will reappear in this list.
+pendingNoSelectedItemError=Please select at least one line.
+
+pendingExpand=expand code
+pendingExpand_tooltip=Expand the source code visualization
+pendingMinimize=minimize code
+pendingMinimize_tooltip=Minimize the source code visualization
+
+noPendingScripts=No pending script approvals.
```
jsoref, line 37:
```suggestion
noPendingScripts=No scripts pending approval.
```
This is my prefered word order.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_approved.jelly
```diff
@@ -0,0 +1,70 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler">
+    <div class="custom-pane-frame-padding">
+        <div>Signatures already approved:</div>
```
jsoref, line 4: Why is some of this not localizable?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_approved.jelly
```diff
@@ -0,0 +1,70 @@
...
+        </textarea>
+        
+        <div class="space-before">Signatures already approved assuming permission check:</div>
+        <textarea readonly="readonly" id="aclApprovedSignatures" rows="10" cols="80">
+            <j:forEach var="line" items="${it.aclApprovedSignatures}">${line}<st:out value="&#10;"/></j:forEach>
+        </textarea>
+        
+        <j:set var="dangerousApprovedSignatures" value="${it.dangerousApprovedSignatures}"/>
+        <j:if test="${!empty(dangerousApprovedSignatures)}">
+            <div class="space-before">Signatures already approved which <strong><font color="red">may have introduced a security vulnerability</font></strong> (recommend clearing):</div>
```
jsoref, line 16:
```suggestion
            <div class="space-before">Signatures already approved which <strong><font color="red">may have introduced security vulnerabilities</font></strong> (recommend clearing):</div>
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_pending.jelly
```diff
@@ -0,0 +1,70 @@
...
+                updateApprovedSignatures(r);
+            });
+        }
+        function clearDangerousApprovedSignatures() {
+            mgr.clearDangerousApprovedSignatures(function(r) {
+                updateApprovedSignatures(r);
+            });
+        }
+    </script>
+</j:jelly>
```
jsoref, line 70:
```suggestion
    </script>
</j:jelly>
```
(Missing newline at eof...)

jeffret-b:
There are some weird, interesting situations with this if multiple jobs have the same script. Some of them probably exist even before this PR but some are exacerbated by the addition of tracking info after the script has been approved.

Here's one example scenario:
1. Create a job `a` with a script that needs to be approved. For example I just used the "Hello World"  example script. 
2. Create a second job `b` with the exact same script that needs to be approved.
3. Login as admin to approve scripts.
4. Only the first version shows as needing approval via the job `a` context.
5. Check job `b`. It no longer shows that the script needs approval.
6. Run job `b`.
7. Check the "Script - Approved" tab. It still shows the context as job `a` even though it has never run.

I'm not sure this is really a problem, though. First, it's rather a corner case for the exact same script to be used in multiple different jobs. I can imagine some situations in which that might be valid, but they're kind of odd and I'm not sure we need to worry about them. Second, even the behavior is odd, I don't know that it is wrong or harmful. The gain from improving the UI / UX here probably outweighs these oddities in these corner cases.

jeffret-b:
A very minor detail at this point, but it looks odd to me to have the tab named "Script - approved" but then to have the content title be "Approved scripts". The first tab for "Script - pending approvals" is more similar to "Scripts pending approval" but still a little disconcerting, especially the shift from singular to plural.

oleg-nenashev:
Thank you for this contribution! It is much appreciated.

Are you doing it as a part of the [Jenkins UI/UX Hackfest](https://www.jenkins.io/events/online-hackfest/2020-uiux/)? If so, please consider reporting this contribution [here](https://github.com/jenkinsci/ui-ux-hackfest-2020/issues/new/choose) so that we could facilitate reviews and share the story 👍

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.jelly
```diff
@@ -0,0 +1,204 @@
...
+                                           data-hash="${hash}"
+                                           class="js-checkbox-hash js-bypass-click" />
+                                </td>
+                                <td class="hash">
+                                    <label for="check-approved-${hash}" title="${hash}" class="js-bypass-click">${hash.substring(0, 8)}</label>
+                                </td>
+                                <j:choose>
+                                    <j:when test="${metadata.isEmpty()}">
+                                        <td colspan="10">
+                                            <i title="${%emptyMetadata_tooltip}">${%emptyMetadata}</i>
```
jglick, line 58: Screenshots in the PR description would be nice!

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/classPath_pending.jelly
Wadeck: It's part of the copy-paste/transfer from the existing script, nothing changed there (yet).

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_approved.jelly
Wadeck: No changes applied in signature / class path parts except code "moved" into tabs layout

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.jelly
Wadeck: Actually you already have one :) Screenshots > "Full script already approved"

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
Wadeck: Any other proposal?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
Wadeck: Screenshots already included in the PR description, the second one contains this button :)

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
Wadeck: In my mind we are listing in the table the different approvals, but saying it's the list of scripts, that are already approved, make actually more sense.

I will keep aside the delete/revoke/remove as it seems we don't have a consensus yet.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_pending.jelly
Wadeck, line 70:
```suggestion
</j:jelly>

```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
Wadeck: By default it's enabled but you can disabled it by either using the system property or running the "reverse" script.

Changing to: "Metadata gathering disabled."

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.jelly
jglick: Sorry, missed the collapsible section.

jglick:
It looks nice. I am just concerned that by making the whole-script approval system look nice, we are implicitly encouraging people to _use_ it, when really this was a terrible idea that I should never have introduced to begin with and it would better be deleted. :man_shrugging:

Wadeck:
> it would better be deleted

If you have a good wording / warning / recommended approach, or similar, please share it/them, I can add them ;)

I was working on this topic because I got multiple feedbacks that the UX was not helpful and some users had 100s of approved scripts without knowing if they are using them or not.

Wadeck:
Added some annotated screenshots and put more highlight on that section

jglick:
> recommended approach

Preferably use the sandbox. Where this is impractical, because your script is basically implementing plugin-like functionality, use a (trusted / global) Groovy library, in the case of Pipeline. For scripting in freestyle projects, if there is no other alternative, limit job configuration to Jenkins administrators (was `RUN_SCRIPTS`, now `ADMINISTER` but not `MANAGE`), who can freely save scripts without them appearing on this page.

Wadeck:
> who can freely save scripts without them appearing on this page.

Are you sure about this part? From my limited testing, when an admin configures a script / job / pipeline, the script is automatically approved, but the hash is still stored. (It was the reason why I introduced the green tick about "preapproved" scripts, to differentiate the two situations)

jglick:
> the script is automatically approved, but the hash is still stored

Right, the script may appear on this page, I just meant that in this scenario there is no reason to ever visit this page.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_approved.jelly
```diff
@@ -0,0 +1,70 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler">
+    <div class="custom-pane-frame-padding">
```
fqueiruga, line 3: This should have the `.pane-frame` before the `.custom-pane-frame-padding` class. Reasons:

- Many styles are applied to `.bigtable` and `.pane` with selectors nested under `.pane-frame`.
- It will make the table styles not be able to be updated when the tables are restyled on Jenkins side.

jsoref:
Fwiw, it'd be nice if i could sort by code, language, requester, request date - (the others are probably meh, but...)

`approve all selected` / `deny all selected` -> please drop `all`, it's superfluous.

I'm not a fan of `expand code` / `minimize code` -> `show code` / `hide code`?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_approved.jelly
```diff
@@ -0,0 +1,70 @@
...
+            <button onclick="if (confirm('Really delete all approvals? Any existing scripts will need to be rerun and signatures reapproved.')) {clearApprovedSignatures()}">Clear Approvals</button>
+        </p>
+        <j:if test="${!empty(dangerousApprovedSignatures)}">
+            Or you can just remove the dangerous ones:
+            <button onclick="clearDangerousApprovedSignatures()">Clear only dangerous Approvals</button>
+        </j:if>
+    </div>
+    
+    <script>
+        var mgr = <st:bind value="${it}"/>;
```
fqueiruga, line 32: XSS risk here. Is it protected?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_pending.jelly
```diff
@@ -0,0 +1,70 @@
...
+                            </j:if>
+                        </p>
+                    </div>
+                </j:forEach>
+            </j:otherwise>
+        </j:choose>
+    </div>
+    
+    <script>
+        var mgr = <st:bind value="${it}"/>;
```
fqueiruga, line 32: Another XSS risk

fqueiruga:
Approving the CSS code (it's namespaced) and JS code (was only moved). I've got a few comments regarding potential XSS vectors. Other than that looks solid

fqueiruga:
One more question. Any chance you can use the same tab templates as provided by Jenkins core? Maybe you could even extend them if you need it. Their colours and borders will probably change in the following weeks and it'd be a shame if this plugin gets left behind. It can always be adapted then but it would be double the work.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_approved.jelly
```diff
@@ -0,0 +1,70 @@
...
+        <j:set var="dangerousApprovedSignatures" value="${it.dangerousApprovedSignatures}"/>
+        <j:if test="${!empty(dangerousApprovedSignatures)}">
+            <div class="space-before">Signatures already approved which <strong><font color="red">may have introduced security vulnerabilities</font></strong> (recommend clearing):</div>
+            <textarea readonly="readonly" id="dangerousApprovedSignatures" rows="10" cols="80">
+                <j:forEach var="line" items="${dangerousApprovedSignatures}">${line}<st:out value="&#10;"/></j:forEach>
+            </textarea>
+        </j:if>
+        <p>
+            You can also remove all previous signature approvals:
+            <button onclick="if (confirm('Really delete all approvals? Any existing scripts will need to be rerun and signatures reapproved.')) {clearApprovedSignatures()}">Clear Approvals</button>
```
fqueiruga, line 23:
```suggestion
            <span class="yui-button">
                <button onclick="if (confirm('Really delete all approvals? Any existing scripts will need to be rerun and signatures reapproved.')) {clearApprovedSignatures()}">Clear Approvals</button>
            </span>
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_pending.jelly
```diff
@@ -0,0 +1,70 @@
...
+            <j:when test="${it.pendingSignatures.isEmpty()}">
+                <p>
+                    No pending signature approvals.
+                </p>
+            </j:when>
+            <j:otherwise>
+                <j:forEach var="s" items="${it.pendingSignatures}">
+                    <div id="s-${s.hash}">
+                        <p>
+                            <button onclick="approveSignature('${s.signature}', '${s.hash}')">Approve</button> /
```
fqueiruga, line 14:
```suggestion
                            <span class="yui-button"><button onclick="approveSignature('${s.signature}', '${s.hash}')">Approve</button></span> /
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_pending.jelly
```diff
@@ -0,0 +1,70 @@
...
+                            <j:if test="${!s.signature.startsWith('field')}">
+                                <button onclick="aclApproveSignature('${s.signature}', '${s.hash}')">Approve assuming permission check</button> /
```
fqueiruga, line 16:
```suggestion
                                <span class="yui-button"><button onclick="aclApproveSignature('${s.signature}', '${s.hash}')">Approve assuming permission check</button></span> /
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_pending.jelly
```diff
@@ -0,0 +1,70 @@
...
+                            </j:if>
+                            <button onclick="denySignature('${s.signature}', '${s.hash}')">Deny</button> signature
```
fqueiruga, line 18:
```suggestion
                            <span class="yui-button"><button onclick="denySignature('${s.signature}', '${s.hash}')">Deny</button></span> signature
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/classPath_pending.jelly
```diff
@@ -0,0 +1,74 @@
...
+                $('pendingClasspathEntries').childElements().each(function(e){e.remove()});
+                $('pendingClasspathEntries').hide();
+            } else {
+                $('pendingClasspathEntries-none').hide();
+                $('pendingClasspathEntries').childElements().each(function(e){e.remove()});
+                /*
+                   Create a list like:
+                    <p id="pcp-${pcp.hash}">
+                        <button class="approve" onclick="approveClasspathEntry('${pcp.hash}')">Approve</button> /
+                        <button class="deny"    onclick="denyClasspathEntry('${pcp.hash}')">Deny</button>
```
fqueiruga, line 26:
```suggestion
                        <span class="yui-button"><button class="approve" onclick="approveClasspathEntry('${pcp.hash}')">Approve</button></span> /
                        <span class="yui-button"><button class="deny"    onclick="denyClasspathEntry('${pcp.hash}')">Deny</button></span> 
```

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/metadata/FullScriptMetadata.java
```diff
@@ -0,0 +1,230 @@
...
+
+@Restricted(NoExternalUse.class)
+public class FullScriptMetadata {
+    
+    public static final FullScriptMetadata EMPTY = new FullScriptMetadata(true);
+
+    /**
+     * In characters
+     */
+    private int scriptLength = -1;
```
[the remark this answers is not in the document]
Wadeck: Removed

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.jelly
Wadeck: Removed

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_approved.jelly
Wadeck: signature_approved = only moved, nothing intented to be changed

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_approved.jelly
```diff
@@ -0,0 +1,70 @@
...
+        </p>
+        <j:if test="${!empty(dangerousApprovedSignatures)}">
+            Or you can just remove the dangerous ones:
+            <button onclick="clearDangerousApprovedSignatures()">Clear only dangerous Approvals</button>
```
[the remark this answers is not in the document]
Wadeck: signature_approved = only moved, nothing intented to be changed

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_pending.jelly
Wadeck: signature_pending = only moved, nothing intented to be changed

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_pending.jelly
Wadeck: signature_pending = only moved, nothing intented to be changed

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/signature_pending.jelly
Wadeck: signature_pending = only moved, nothing intented to be changed

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/classPath_pending.jelly
Wadeck: classPath_pending = only moved, nothing intented to be changed

Wadeck:
> Was this intented?

Yes, the code that was just split/moved is not modified.

> Any chance you can use the same tab templates as provided by Jenkins core? Maybe you could even extend them if you need it

Due to the lack of extensibility of tags in general, I was forced to copy-paste the stuff over to get the information badge in the tab headers that are providing values about what requires a look. If this change is done in core, this plugin could use that, but in the meantime, the plugin should use a "fork" otherwise that information is missing.
If you know a way to extend the existing without rewriting, let me know!

> XSS risk here. Is it protected?

We are not aware of any XSS risk by using `<st:bind value="${it}"/>;`, but be careful as the code was just moved, if it was a real XSS, it means that you disclosed it in this public PR. The only XSS that was discovered during this work was corrected as SECURITY-1866.

For your information, the result of that call gives (with some random values):
```
var mgr = makeStaplerProxy('/jenkins/$stapler/bound/b4496557-xxxx-xxxx-xxxx-ee4d8b251c51','xxxxdea5xxxxe0fd15b23b28e66913e476e7bff91f0e936e307a0002f12c6bcf',['denyApprovedClasspathEntry','aclApproveSignature','denyClasspathEntry','clearDangerousApprovedSignatures','approveClasspathEntry','clearApprovedClasspathEntries','clearApprovedSignatures','clearApprovedScripts','getClasspathRenderInfo','approveSignature','denySignature']);
```

> .pane-frame related CSS issue

Unfortunately it's very complicated to have a code that is compatible with versions before the table revamp and after it. So I moved "full speed" to be compatible with the version after table revamp. Meaning: removal of the titles and pane-frame (and my custom one) to have table in full size on the panel, like in update-center.

<details>

<summary>Screenshots</summary>

## Before table revamp
### Scripts pending
before_table_script_pending


### Scripts approved
before_table_scripts_approved

### Signatures pending
before_table_signature_pending


## After table revamp

### Scripts pending
after_table_script_pending


### Scripts approved
after_table_script_approved


### Signatures pending
after_table_signature_pending


</details>

Wadeck:
> Fwiw, it'd be nice if i could sort by code, language, requester, request date

Follow-up PR welcome :) 

> approve all selected / deny all selected -> please drop all, it's superfluous. I'm not a fan of expand code / minimize code -> show code / hide code?

Adjusted

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -25,7 +19,8 @@ lastKnownApproverLogin=Last approver
 
 ## table content
 noRequesterSinceMetadata=N/A
-noRequesterSinceMetadata_tooltip=The requester was not recorded. It could be due to the request being created before the metadata introduction or the code calling this script did not provide user information.
+noRequesterSinceMetadata_tooltip=The requester was not recorded. \
+  It could be due to the request being created before the metadata introduction or the code calling this script did not provide user information.
```
jsoref, line 23:
```suggestion
  It could be because the request was created before the metadata was introduced or the code calling this script did not provide user information.
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_pending.properties
```diff
@@ -1,21 +1,16 @@
...
 contextUser=Requester
 contextItem=Context
 requestApprovalTime=Request date
 requestApprovalTime_tooltip=Inform about when the approval request was made.
 
 ## table content
 noRequesterSinceMetadata=N/A
-noRequesterSinceMetadata_tooltip=The requester was not recorded. It could be due to the request being created before the metadata introduction or the code calling this script did not provide user information.
+noRequesterSinceMetadata_tooltip=The requester was not recorded. \
+  It could be due to the request being created before the metadata introduction or the code calling this script did not provide user information.
```
jsoref, line 13:
```suggestion
  It could be because the request was created before the metadata introduced or the code calling this script did not provide user information.
```

src/test/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest/legacyApproval/scriptApproval.xml
```diff
@@ -0,0 +1,84 @@
...
+    <pendingClasspathEntry>
+      <context>
+        <user>config</user>
+        <item>free-for-classpath</item>
+      </context>
+      <hash>7f014e0dab147d3ae431efa1b8b1305112711b18</hash>
+      <url>https://repo.jenkins-ci.org/javanet2-cache/org/jvnet/hudson/main/maven-plugin/1.302/maven-plugin-1.302.hpi</url>
+    </pendingClasspathEntry>
+  </pendingClasspathEntries>
+</scriptApproval>
```
jsoref, line 84: newline?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_pending.properties
daniel-beck: > before the metadata introduced

Ungrammatical?

jsoref: Note to self: copy+paste is my friend. Fixed.

Wadeck:
From Daniel:

1. Create pipeline as 'user'
2. Write a script and do not use sandbox, save.
3. Try to build, it fails
4. As 'admin', approve the script.
5. As 'user' build the job, it works
6. As 'admin', delete from approvals
7. As 'user', build the job, it fails

Expected: Script is now pending
Actual: It's not

daniel-beck:
Some UI feedback:

* I don't think it's useful to show badges with the number of already approved entries. I associated badges with "unread messages", while these tabs don't have content that needs attention. Badges should be limited to content requiring action. 
* I wonder whether it makes sense to hide tabs for empty lists, so that the most common UI would only show "approved scripts" and "approved signatures", and classpath-related content to only show rarely. This could probably go either way.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,59 @@
...
+## table content
+noRequesterSinceMetadata=N/A
+noRequesterSinceMetadata_tooltip=The requester was not recorded. \
+  It could be because the request was created before the metadata was introduced or the code calling this script did not provide user information.
+noRequestApprovalTimeSinceMetadata=N/A
+noRequestApprovalTimeSinceMetadata_tooltip=It was requested before the introduction of metadata.
+noContextItemSinceMetadata=N/A
+noContextItemSinceMetadata_tooltip=The context item was not provided by the code asking for approval.
+
+emptyMetadata=There is no metadata for this approval
```
daniel-beck, line 29:
```suggestion
emptyMetadata=This script has been approved before Script Security 1.75 and no additional metadata has been recorded yet. Once this script is executed, additional metadata will be added.
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,59 @@
...
+emptyMetadata_tooltip=If the script is used after the metadata introduction, some metadata will be displayed here. \
```
daniel-beck, line 30: This tooltip lacks discoverability.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,59 @@
...
+notUsedSinceMetadata=N/A
+notUsedSinceMetadata_tooltip=It was not used since the introduction of metadata.
+notApprovedSinceMetadata=N/A
+notApprovedSinceMetadata_tooltip=It was not approved since the introduction of metadata.
+noKnownApproverSinceMetadata=N/A
+noKnownApproverSinceMetadata_tooltip=It was not approved since the introduction of metadata.
+
+revokeApproval=Remove selected
+revokeApproval_tooltip=Remove an approved script can have an impact on the jobs using it. \
+  The next time an attempt to execute the script is made, a new approval will be required.
```
daniel-beck, line 49:
```suggestion
revokeApproval_tooltip=Removing a script from this list will revoke approval for its execution. \
  The next attempt to execute the script will fail, and the script will be added to the list of scripts requiring approval.
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,59 @@
...
+revokeApproval_confirm=Really delete all selected approvals? Any existing scripts will need to be requeued and reapproved.
```
daniel-beck, line 50:
```suggestion
revokeApproval_confirm=Really revoke approvals for all selected scripts? All scripts not using the sandbox will need to be approved again before they can be executed successfully, and will fail until they are.
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,59 @@
...
+wasPreapproved_tooltip=The script content was approved directly during configuration, meaning it comes from a user with the permission to run scripts.
+wasPreapproved_alt=Preapproved
+noUsageCountSinceMetadata=N/A
+noUsageCountSinceMetadata_tooltip=It was not used since the introduction of metadata.
+notUsedSinceMetadata=N/A
+notUsedSinceMetadata_tooltip=It was not used since the introduction of metadata.
+notApprovedSinceMetadata=N/A
+notApprovedSinceMetadata_tooltip=It was not approved since the introduction of metadata.
+noKnownApproverSinceMetadata=N/A
+noKnownApproverSinceMetadata_tooltip=It was not approved since the introduction of metadata.
```
daniel-beck, line 45: When are these shown?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,59 @@
...
+  <p>You can re-enable it permanently by removing the system property: <br />\
+  <code>org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval.metadataGathering</code> </p>\
+  <p>or temporarily (until restart) by running this in the Script Console: <br />\
+  <code>org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval.METADATA_GATHERING = true</code></p>
+
+## table headers
+approvedExpandCode=Code
+approvedLanguage=Language
+usageCount=# of uses
+usageCount_tooltip=Uses since the introduction of metadata
```
daniel-beck, line 12:
```suggestion
usageCount_tooltip=How often this script was executed. Only executions since Script Security plugin was updated to version 1.75.
```

daniel-beck:
Haven't looked in-depth at the code yet, but this looks a lot like a change that necessitates a compatibility warning because downgrades will be difficult. Thoughts?

Wadeck:
> I don't think it's useful to show badges with the number of already approved entries. I associated badges with "unread messages", while these tabs don't have content that needs attention. Badges should be limited to content requiring action.
Great idea, I will change this.

> I wonder whether it makes sense to hide tabs for empty lists, so that the most common UI would only show "approved scripts" and "approved signatures", and classpath-related content to only show rarely. This could probably go either way.
I am not fan of UI elements being hidden until required. This reduce the likelihood the users will know where to go when they need this. If they are "used" to have them shown, it's easier.

> Haven't looked in-depth at the code yet, but this looks a lot like a change that necessitates a compatibility warning because downgrades will be difficult. Thoughts?
I am just adding metadata. If you go back, the metadata will not be used any longer. The original XML file is not touched.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
jsoref:
```suggestion
emptyMetadata=This script was approved before Script Security 1.75 and no additional metadata has been recorded yet. Once this script is executed, additional metadata will be added.
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
jsoref: Fwiw while I suggest changing a `has been` elsewhere to `was`, these should be `has not been`.

daniel-beck:
Looking at something like

> Screenshot

I wonder whether it would make sense to use a checksum prefix to identify the specific entry, both when there is, and when there is not, additional metadata. The checksum is all we have in some cases, but we always have that. 

> Screenshot 2020-08-24 at 20 27 01

(I'm not proposing to remove the pre-approval checkmark, but Preview.app just isn't a good drawing program!)

jglick:
Maybe we can merge something that is good enough? This entire PR touches on functionality which you should never be using to begin with, so unless the behavior is clearly wrong, it does not make sense to sink a lot of time into it IMO.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -751,24 +981,143 @@ public ApprovedWhitelist() {
...
+                    String script = pendingScript == null ? null : pendingScript.script;
+                    this.metadataStorage.withMetadata(hash, script, metadata ->
+                            metadata.notifyApproval(pendingScript, userLogin)
+                    );
+                }
+            }
+
+            save();
+        }
+        SecurityContext orig = ACL.impersonate(ACL.SYSTEM);
```
jglick, line 1051: This is deprecated! Use `ACL.as`, and fix the existing two usages in this class while you are at it.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -943,4 +1292,21 @@ public synchronized JSON clearApprovedClasspathEntries() throws IOException {
...
+     * To have an effectively final variable
+     */
+    private @CheckForNull String getCurrentUserLogin() {
+        String userLogin = null;
+        Authentication auth = SecurityContextHolder.getContext().getAuthentication();
+        if (auth != null) {
+            userLogin = auth.getName();
+        }
+
+        return userLogin;
```
jglick, line 1305: or

```suggestion
        Authentication auth = Jenkins.getAuthentication();
        return auth.equals(Jenkins.ANONYMOUS) ? null : auth.getName();
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
Wadeck: When a script is approved before the metadata introduction, and then is used. We do not have the information on who approved it.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
Wadeck: Both proposals seem to pollute the space due to the size of the sentence.

<details>

<summary>Proof</summary>

### After
Screenshot_2020-09-02_083609_001

### Before (with the icon for tooltip discoverability)
Screenshot_2020-09-02_083719_001

### Before (with just the cursor: help)
Screenshot_2020-09-02_084251_001

</details>

Going with no modification. This could be improved in a follow up PR.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
Wadeck: I tried with a help icon but it will require to have newer version of core to have the latest help icon, so keeping it simple for the moment with `cursor: help`

<details>

<summary>Cursor</summary>

Screenshot_2020-09-02_084251_001


</details>

Wadeck:
> I wonder whether it makes sense to hide tabs for empty lists, so that the most common UI would only show "approved scripts" and "approved signatures", and classpath-related content to only show rarely. This could probably go either way.

<details>

<summary>New version</summary>

Screenshot_2020-09-02_085343_001


----

Screenshot_2020-09-02_085350_001



</details>

---- 

> From Daniel:
As 'user', [...]
Expected: Script is now pending approval
Actual: It's neither approved nor pending

It's a bug already present before this PR :(

https://github.com/jenkinsci/script-security-plugin/blob/fcb41b7e540a6e0a53f89098e1fb6cc5210195a3/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java#L637

^- in the "using" method.

I will not modify this method as it could have some impacts I am not aware of (like having script being auto-approved or related stuff)

daniel-beck:
> I will not modify this method as it could have some impacts I am not aware of (like having script being auto-approved or related stuff)

AFAIUI this PR newly adds the capability to un-approve specific single scripts, so what wasn't a (big) problem before is now probably much more relevant.

Wadeck:
@dwnusbaum @jglick any thoughts on the recent comments?

Modifying the `using` method could have terrible side effects depending on the plugin calling it. I am not confident enough to propose such bug correction in this PR.

jeffret-b:
> I am not fan of UI elements being hidden until required. This reduce the likelihood the users will know where to go when they need this. If they are "used" to have them shown, it's easier.

I strongly agree with this.

> AFAIUI this PR newly adds the capability to un-approve specific single scripts, so what wasn't a (big) problem before is now probably much more relevant.

This seems like a legitimate concern. Providing a new capability that introduces problems because of existing limitations, creates a valid concern. I'm just not sure how big of a concern it is. Does it override Jesse's earlier desire to "merge something that is good enough"? My tentative answer is that we can merge something that is good enough when it comes to some of the UI and layout, but the introduction of a new, potentially confusing or limiting capability needs correction.

Wadeck:
Due to the bug discovered by Daniel, this PR is currently blocked by [JENKINS-63668](https://issues.jenkins-ci.org/browse/JENKINS-63668).

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,58 @@
...
+noContextItemSinceMetadata_tooltip=The context item was not provided by the code asking for approval.
+
+emptyMetadata=There is no metadata for this approval.
+emptyMetadata_tooltip=If the script is used after the metadata introduction, some metadata will be displayed here. \
+  It means that an unused approval will not have metadata and could be reasonably removed after a certain period of time.
+noScriptCodeSinceMetadata=N/A
+noScriptCodeSinceMetadata_tooltip=The approval was given while the metadata was not gathered.
+approvedContextUser=Requester
+approvedContextItem=Context
+wasPreapproved_tooltip=The script content was approved directly during configuration, meaning it comes from a user with the permission to run scripts.
```
jsoref, line 34:
```suggestion
wasPreapproved_tooltip=The script content written by a user with permission to run scripts (implicit approval).
```

src/main/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval/tabs/fullScript_approved.properties
```diff
@@ -0,0 +1,58 @@
...
+wasPreapproved_alt=Preapproved
+noUsageCountSinceMetadata=N/A
+noUsageCountSinceMetadata_tooltip=It was not used since the introduction of metadata.
+notUsedSinceMetadata=N/A
+notUsedSinceMetadata_tooltip=It was not used since the introduction of metadata.
+notApprovedSinceMetadata=N/A
+notApprovedSinceMetadata_tooltip=It was not approved since the introduction of metadata.
+noKnownApproverSinceMetadata=N/A
+noKnownApproverSinceMetadata_tooltip=It was not approved since the introduction of metadata.
```
jsoref, line 43:
```suggestion
noUsageCountSinceMetadata_tooltip=It has not been used since the introduction of metadata.
notUsedSinceMetadata=N/A
notUsedSinceMetadata_tooltip=It has not been used since the introduction of metadata.
notApprovedSinceMetadata=N/A
notApprovedSinceMetadata_tooltip=It has not been approved since the introduction of metadata.
noKnownApproverSinceMetadata=N/A
noKnownApproverSinceMetadata_tooltip=It has not been approved since the introduction of metadata.
```

fqueiruga:
How can we resurrect this?

Wadeck:
@fqueiruga Either you convince the people (mainly Jeff, a bit Daniel) blocking this PR that the existing bug should not block this PR, or convince pipeline people the reported bug should be prioritized :)

PR #304, opened 2020-07-23
[JENKINS-63171] Add more generic whitelist entries

jvz:
https://issues.jenkins-ci.org/browse/JENKINS-63171

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -642,6 +642,7 @@ staticMethod org.codehaus.groovy.runtime.DateGroovyMethods upto java.util.Calend
 staticMethod org.codehaus.groovy.runtime.DateGroovyMethods upto java.util.Date java.util.Date groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods abs java.lang.Number
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods addAll java.util.Collection java.lang.Object[]
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods and java.lang.Boolean java.lang.Boolean
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods any java.lang.Iterable groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods any java.lang.Object groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods any java.util.Map groovy.lang.Closure
@@ -965,7 +966,9 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods take java.util.Lis
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods takeRight java.lang.Iterable int
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods takeRight java.util.List int
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods times java.lang.Number groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toBoolean java.lang.Boolean
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toBoolean java.lang.String
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toDouble java.lang.String
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toInteger java.lang.String
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toList boolean[]
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toList byte[]
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -1118,6 +1118,7 @@ staticMethod org.codehaus.groovy.runtime.StringGroovyMethods reverse java.lang.S
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods takeWhile groovy.lang.GString groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods takeWhile java.lang.CharSequence groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods toBoolean java.lang.String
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -642,6 +642,7 @@ staticMethod org.codehaus.groovy.runtime.DateGroovyMethods upto java.util.Calend
 staticMethod org.codehaus.groovy.runtime.DateGroovyMethods upto java.util.Date java.util.Date groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods abs java.lang.Number
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods addAll java.util.Collection java.lang.Object[]
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods and java.lang.Boolean java.lang.Boolean
```
dwnusbaum, line 645: Definitely safe: https://github.com/apache/groovy/blob/731d4daa78b9cd32bea724d2d651aa1164eef6cc/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java#L15348.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -965,7 +966,9 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods take java.util.Lis
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods takeRight java.lang.Iterable int
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods takeRight java.util.List int
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods times java.lang.Number groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toBoolean java.lang.Boolean
```
dwnusbaum, line 969: Looks like this method is `@Deprecated` and delegates to `StringGroovyMethods.toBoolean`, so I would also add that method at the same time. The implementation is safe, although a bit weird (special cases for `"y"` and `"1"`): https://github.com/apache/groovy/blob/731d4daa78b9cd32bea724d2d651aa1164eef6cc/src/main/org/codehaus/groovy/runtime/StringGroovyMethods.java#L3240-L3248

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -965,7 +966,9 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods take java.util.Lis
...
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toBoolean java.lang.String
+staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethods toDouble java.lang.String
```
dwnusbaum, line 971: Ditto, this method is `@Deprecated` and delegates to `StringGroovyMethods.toDouble(CharSequence)`, so I would add that method at the same time. The implementation looks safe: https://github.com/apache/groovy/blob/731d4daa78b9cd32bea724d2d651aa1164eef6cc/src/main/org/codehaus/groovy/runtime/StringGroovyMethods.java#L3270-L3272

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1114,6 +1117,9 @@ staticMethod org.codehaus.groovy.runtime.StringGroovyMethods replaceFirst java.l
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods reverse java.lang.String
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods takeWhile groovy.lang.GString groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods takeWhile java.lang.CharSequence groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.StringGroovyMethods toBoolean java.lang.Boolean
+staticMethod org.codehaus.groovy.runtime.StringGroovyMethods toDouble java.lang.Double
+staticMethod org.codehaus.groovy.runtime.StringGroovyMethods toInteger java.lang.Integer
```
dwnusbaum, line 1122: This isn't a real method, did you mean `StringGroovyMethods.toInteger(String)` and the nondeprecated version `StringGroovyMethods.toInteger(CharSequence)`?

jvz: Oops, that's what I get for manually typing it out!

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1117,9 +1117,10 @@ staticMethod org.codehaus.groovy.runtime.StringGroovyMethods replaceFirst java.l
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods reverse java.lang.String
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods takeWhile groovy.lang.GString groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods takeWhile java.lang.CharSequence groovy.lang.Closure
-staticMethod org.codehaus.groovy.runtime.StringGroovyMethods toBoolean java.lang.Boolean
-staticMethod org.codehaus.groovy.runtime.StringGroovyMethods toDouble java.lang.Double
-staticMethod org.codehaus.groovy.runtime.StringGroovyMethods toInteger java.lang.Integer
+staticMethod org.codehaus.groovy.runtime.StringGroovyMethods toBoolean java.lang.String
+staticMethod org.codehaus.groovy.runtime.StringGroovyMethods toDouble java.lang.String
+staticMethod org.codehaus.groovy.runtime.StringGroovyMethods toInteger java.lang.CharSequence
+staticMethod org.codehaus.groovy.runtime.StringGroovyMethods toInteger java.lang.String
```
dwnusbaum, line 1124: Ok one last thing, but `DefaultGroovyMethods.toInteger(String)` (which is deprecated) is (I think) the method that `SandboxInterceptor` will actually think you are calling if you write something like `"123".toInteger()`, so you probably need to add that one too. Regular Groovy does not consider deprecated methods when selecting extension methods, so I asked you to add the non-deprecated `StringGroovyMethods` equivalents in case we ever fix that bug 😅

jvz: I still have that in the file, don't I?

dwnusbaum: Ah, I missed that it was already present before your changes, so yeah it should be fine as-is.

PR #305, opened 2020-07-27
[JENKINS-63218] Add a script approval endpoint for Groovy Scripts

PierreBtz:
Some details I'm not sure about:
1. Regarding the return code, as I'm technically creating a resource I should probably return an http created code, but it would also mean providing a location, I'm not 100% clear what that should be. Therefore I chose to return an http ok for now.
2. I'm not 100% sure about what the checkScriptForCompilationErrors is exactly catching, I made some obviously compilation error and it wasn't caught, therefore I couldn't manually test this use case.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
@@ -24,6 +24,8 @@
 
 package org.jenkinsci.plugins.scriptsecurity.scripts;
 
+import groovy.lang.GroovyClassLoader;
+import hudson.model.Api;
 import jenkins.model.GlobalConfiguration;
 import jenkins.model.GlobalConfigurationCategory;
 import net.sf.json.JSONArray;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -943,4 +956,36 @@ public synchronized JSON clearApprovedClasspathEntries() throws IOException {
...
+
+        try(BufferedReader reader =
+                  new BufferedReader(new InputStreamReader(request.getInputStream(), StandardCharsets.UTF_8))) {
+            String script = reader
+                  .lines()
+                  .collect(Collectors.joining("\n"));
+            FormValidation scriptValidation = GroovySandbox.checkScriptForCompilationErrors(script,
+                  new GroovyClassLoader(Jenkins.get().getPluginManager().uberClassLoader));
+            if (scriptValidation.kind != FormValidation.Kind.OK) {
+                return scriptValidation;
```
daniel-beck, line 985: This doesn't look like success as documented below.

dwnusbaum (changes requested):
Discussed with PR author separately, but I really would prefer to not introduce an API for full-script approval. I think it is likely to be used unsafely, and in general users should be using the sandbox instead. At the very least, if we add an API for full-script approval I think we should add an API for approving signatures for use in the sandbox at the same time so we do not mislead users into thinking full-script approval is preferred since it has better API support.

That said, instead of requiring administrators to make complex decisions about safety of various scripts or signatures, I think it would be better to spend some time adding telemetry to signature approvals and then use the telemetry to update `generic-whitelist` and `blacklist` (maybe add a new `safe-but-use-sh` list?) with all of the signatures we see coming in from real instances. Once we do that, we should update documentation, especially to list examples of issues with signatures in `blacklist`  (e.g. weird syntax leading to request to approve `GroovyObject.invokeMethod`, reasons why various methods are unsafe) to help users understand what they should be doing instead, so that over time admins would have to interact with the sandbox less frequently, in which case APIs like this are not as relevant.

PierreBtz:
Closing as I won't be pushing this farther, approach proposed by Devin seems better.

PR #309, opened 2020-09-25
Allow some additional String and Base64 related methods

daniel-beck:
Untested.


dwnusbaum (approved):
Looks fine to me. I refiled https://github.com/jenkinsci/script-security-plugin/pull/310 on top of this PR to add almost all of the remaining signatures for `String`.

PR #310, opened 2020-09-28
Allow some additional String and Base64 related methods

dwnusbaum:
Subsumes https://github.com/jenkinsci/script-security-plugin/pull/309 and adds every remaining `String`-related signature other than `String.intern`, which while not necessarily unsafe, seems a bit questionable.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -72,6 +72,7 @@ staticMethod java.lang.Boolean valueOf boolean
 staticMethod java.lang.Boolean valueOf java.lang.String
 method java.lang.CharSequence charAt int
 method java.lang.CharSequence length
+method java.lang.CharSequence subSequence int int
 method java.lang.Class getName
 method java.lang.Class getSimpleName
 method java.lang.Class isInstance java.lang.Object
@@ -111,6 +112,8 @@ method java.lang.String endsWith java.lang.String
 method java.lang.String equalsIgnoreCase java.lang.String
 staticMethod java.lang.String format java.lang.String java.lang.Object[]
 staticMethod java.lang.String format java.util.Locale java.lang.String java.lang.Object[]
+method java.lang.String getBytes
+method java.lang.String getBytes java.lang.String
 method java.lang.String indexOf int
 method java.lang.String indexOf int int
 method java.lang.String indexOf java.lang.String
@@ -121,6 +124,8 @@ method java.lang.String lastIndexOf int int
 method java.lang.String lastIndexOf java.lang.String
 method java.lang.String lastIndexOf java.lang.String int
 method java.lang.String matches java.lang.String
+method java.lang.String regionMatches boolean int java.lang.String int int
+method java.lang.String regionMatches int java.lang.String int int
 method java.lang.String replace char char
 method java.lang.String replace java.lang.CharSequence java.lang.CharSequence
 method java.lang.String replaceAll java.lang.String java.lang.String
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -107,23 +107,46 @@ method java.lang.Object equals java.lang.Object
...
 method java.lang.Object hashCode
 method java.lang.Object toString
+new java.lang.String
+staticField java.lang.String CASE_INSENSITIVE_ORDER
+new java.lang.String byte[]
+new java.lang.String byte[] int int
+new java.lang.String byte[] int int java.lang.String
+new java.lang.String byte[] int int java.nio.charset.Charset
+new java.lang.String byte[] java.lang.String
+new java.lang.String byte[] java.nio.charset.Charset
```
daniel-beck, line 117: I left this out, because `Charset` isn't otherwise allowed.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -107,23 +107,46 @@ method java.lang.Object equals java.lang.Object
...
+new java.lang.String char[]
+new java.lang.String char[] int int
+method java.lang.String codePointAt int
+method java.lang.String codePointBefore int
+method java.lang.String concat java.lang.String
 method java.lang.String contains java.lang.CharSequence
+method java.lang.String contentEquals java.lang.CharSequence
+method java.lang.String contentEquals java.lang.StringBuffer
```
daniel-beck, line 125: Similarly, `StringBuffer` doesn't seem to be allowed? Why add this then?

daniel-beck (commented):
Thanks for the completion.

I checked a Javadoc and thought everything was there. Did I accidentally look at Java 5?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
dwnusbaum: Yeah good point, I can add the fields in `StandardCharsets`

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
dwnusbaum: I just did not notice, but I can add methods related to it (same for `StringBuilder`). In general, leaving out methods where operations related to the argument types are not yet whitelisted gets confusing, since we have to remember to go back and add those methods whenever we add whitelist entries for new types, so I just try to add all safe methods defined on a receiver at the same time, even if we haven't whitelisted methods related to the argument types.

car-roll (approved):

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
dwnusbaum: I added the `StandardCharsets` fields and `Charset.forName` in https://github.com/jenkinsci/script-security-plugin/pull/310/commits/ce20e20786fe950959827fddf3b53234bca302f4.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
dwnusbaum: I added methods defined in `AbstractStringBuilder` and constructors for `StringBuilder` and `StringBuffer` in https://github.com/jenkinsci/script-security-plugin/pull/310/commits/abd7cd362ba160851d0d33b9328a02caa5ee7b37. There is no reason to ever use `StringBuffer` instead of `StringBuilder` in the sandbox, but since it is safe, I think it makes sense to include it here anyway in case someone happens to use it.

daniel-beck (approved):

PR #312, opened 2020-10-09
Add Dependabot configuration

dwnusbaum:


jglick (approved):

timja (approved):

olamy (approved):
LGTM even if weekly is maybe better to avoid too much noise :)

dwnusbaum:
@olamy This plugin only has a few dependencies, so let's see how much noise there is with daily updates and then we can switch to weekly if needed.

PR #319, opened 2020-10-28
JENKINS-64057: Whitelist XmlUtil and a few java.util scripts

dasanjan1296:
This would whitelist the following signatures:

- staticMethod groovy.xml.XmlUtil escapeXml java.lang.String
- staticField java.util.concurrent.TimeUnit MILLISECONDS
- staticField java.util.concurrent.TimeUnit SECONDS
- new java.util.TreeMap

All of these signatures are utility classes and hence, it should be safe to whitelist them.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -12,6 +12,7 @@ staticMethod groovy.json.JsonOutput toJson java.util.Calendar
 staticMethod groovy.json.JsonOutput toJson java.util.Date
 staticMethod groovy.json.JsonOutput toJson java.util.Map
 staticMethod groovy.json.JsonOutput toJson java.util.UUID
+staticMethod groovy.xml.XmlUtil escapeXml java.lang.String
 new groovy.json.JsonSlurper
 method groovy.json.JsonSlurper parseText java.lang.String
 
@@ -478,6 +479,7 @@ staticField java.time.format.DateTimeFormatter ISO_ZONED_DATE_TIME
 staticField java.time.format.DateTimeFormatter RFC_1123_DATE_TIME
 staticMethod java.time.format.DateTimeFormatter ofPattern java.lang.String
 new java.util.ArrayList java.util.Collection
+new java.util.concurrent.LinkedBlockingQueue
 staticMethod java.util.Arrays asList java.lang.Object[]
 staticMethod java.util.Arrays toString java.lang.Object[]
 staticMethod java.util.Base64 getDecoder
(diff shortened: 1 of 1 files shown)


dasanjan1296:
@dwnusbaum Can you please take a look? This has been pending for a long time. Thanks!

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -738,11 +739,16 @@ method java.util.Random nextInt
 method java.util.Random nextInt int
 method java.util.Random nextLong
 staticMethod java.util.TimeZone getTimeZone java.lang.String
+new java.util.TreeMap
 new java.util.TreeSet
 staticMethod java.util.UUID randomUUID
+method java.util.concurrent.BlockingQueue poll long java.util.concurrent.TimeUnit
 method java.util.concurrent.Callable call
+new java.util.concurrent.LinkedBlockingQueue
```
dwnusbaum, line 747: I'm not sure if it makes sense to add the `BlockingQueue`-related signatures for two main reasons:
* If you use blocking methods like `BlockingQueue.poll` in a Pipeline, the CPS VM thread will be stuck, which will prevent the Pipeline from executing correctly (for example other parallel branches will not be able to progress).
* These signatures seem indicative of using multiple threads from sandboxed scripts, which is a security risk because other threads will execute outside of the sandbox.
 
If you just want a data structure that has queue-like behavior, maybe we could add signatures related to `LinkedList` instead?

dwnusbaum (commented):
@dasanjan1296 Thanks for the PR! I think these signatures are all safe, but I'm not sure if it makes sense to add the `BlockingQueue`-related signatures. See my comment below.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
dasanjan1296: @gvprathyusha6 should be able to clarify more on this.

haridsv: > These signatures seem indicative of using multiple threads from sandboxed scripts, which is a security risk because other threads will execute outside of the sandbox.

@dwnusbaum I would like to better understand the above concern. What do you mean by "other threads"? We are looking at using one `BlockingQueue` object across multiple `parallel` (as in the Jenkins step) branches. We are not explicitly spawning any threads (we can't anyway, inside the sandbo) and we will not be able to expose this object to an outside thread too. Do you see any concern with this approach?

dwnusbaum: @haridsv I am referring to any code in a Pipeline that spawns Java threads. The `parallel` step does not actually spawn multiple Java threads that run in parallel. See the final paragraph [here](https://github.com/jenkinsci/workflow-cps-plugin#technical-design) for more details, but essentially there is only ever one thread executing the Pipeline at a time, so depending on your exact use case you would probably be fine with a regular list or queue instead of needing a `BlockingQueue` since all of the Groovy code in the parallel branches are effectively synchronized with respect to each other.

haridsv: @dwnusbaum Thank you so much, that makes total sense!

dasanjan1296: @dwnusbaum Per our discussion, I have removed the signatures related to BlockingQueue owing to the suggested alternatives. Please take a look.

dwnusbaum (approved):

car-roll (approved):

dasanjan1296:
@dwnusbaum Can this PR be merged now?

PR #325, opened 2020-12-09
Malformed signatures will no longer be added to approved signatures

St1gven:
Adding a malformed signature causes hanging of "In-process Script Approval " like https://issues.jenkins.io/browse/JENKINS-46764.
To avoid this, malformed signatures will no longer be added to the approved list.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
@@ -707,7 +707,7 @@ public class ScriptApproval extends GlobalConfiguration implements RootAction {
             try {
                 reconfigure();
             } catch (IOException e) {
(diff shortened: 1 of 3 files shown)


jglick:
Amending #170?

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -796,17 +796,31 @@ private synchronized void removePendingScript(String hash) {
         Jenkins.getInstance().checkPermission(Jenkins.RUN_SCRIPTS);
         pendingSignatures.remove(new PendingSignature(signature, false, ApprovalContext.create()));
         approvedSignatures.add(signature);
-        save();
-        return reconfigure();
+        try {
+            return reconfigure();
+        } catch (IOException e) {
+            approvedSignatures.remove(signature);
```
jglick, line 802: This does not look right. Any format checks should be done earlier.

PR #329, opened 2021-02-15
[JENKINS-62708] Add JCasC support for approved script hashes

MarkRx:
Adds the ability to manage approved script hashes with JCasC. With the current UX to determine the hash you can either approve it and see what was added to the scriptApproval.xml file on disk or compute the SHA-1 of the script. There is an open effort to approve the UX with [JENKINS-62448](https://issues.jenkins.io/browse/JENKINS-62448) https://github.com/jenkinsci/script-security-plugin/pull/300.

Example:
Script:
`println(jenkins.model.Jenkins.instance.getLabels())`

```
security:
  scriptApproval:
    approvedScriptHashes:
      - fccae58c5762bdd15daca97318e9d74333203106 -- print all labels
```

See [JENKINS-62708](https://issues.jenkins.io/browse/JENKINS-62708)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
@@ -66,6 +66,7 @@ import java.util.TreeSet;
 import java.util.function.Consumer;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import jenkins.model.Jenkins;
@@ -699,6 +700,27 @@ public class ScriptApproval extends GlobalConfiguration implements RootAction {
     @Restricted(NoExternalUse.class) // Jelly, implementation
     public synchronized String[] getAclApprovedSignatures() {
         return aclApprovedSignatures.toArray(new String[aclApprovedSignatures.size()]);
+    }
+
+    @DataBoundSetter
+    public synchronized void setApprovedScriptHashes(String[] scriptHashes) throws IOException {
+        Jenkins.getInstance().checkPermission(Jenkins.RUN_SCRIPTS);
+        approvedScriptHashes.clear();
+        Pattern sha1Pattern = Pattern.compile("[a-fA-F0-9]{40}");
+        for (String scriptHash : scriptHashes) {
+            if (scriptHash != null && sha1Pattern.matcher(scriptHash).matches()) {
+                approvedScriptHashes.add(scriptHash);
+            } else {
+                LOG.warning(() -> "Ignoring malformed script hash: " + scriptHash);
+            }
+        }
+        save();
+        reconfigure();
+    }
+
+    @Restricted(NoExternalUse.class) // Jelly, implementation
+    public synchronized String[] getApprovedScriptHashes() {
+        return approvedScriptHashes.toArray(new String[approvedScriptHashes.size()]);
     }
 
     @Restricted(NoExternalUse.class) // implementation
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/JcascTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/JcascTest.java
@@ -26,6 +26,9 @@ public class JcascTest {
         String[] approved = ScriptApproval.get().getApprovedSignatures();
         assertTrue(approved.length == 1);
         assertEquals(approved[0], "method java.net.URI getHost");
+        String[] approvedScriptHashes = ScriptApproval.get().getApprovedScriptHashes();
+        assertTrue(approvedScriptHashes.length == 1);
+        assertEquals(approvedScriptHashes[0], "fccae58c5762bdd15daca97318e9d74333203106");
     }
 
     @Test
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest.java
@@ -162,15 +162,20 @@ public class ScriptApprovalTest extends AbstractApprovalTest<ScriptApprovalTest.
         assertEquals(0, sa.getDangerousApprovedSignatures().length);
     }
 
-    @Issue("JENKINS-57563")
-    @LocalData // Just a scriptApproval.xml that whitelists 'staticMethod jenkins.model.Jenkins getInstance'
+    @Issue({"JENKINS-57563", "JENKINS-62708"})
+    @LocalData // Just a scriptApproval.xml that whitelists 'staticMethod jenkins.model.Jenkins getInstance' and a script printing all labels
     @Test
     public void upgradeSmokes() throws Exception {
+        configureSecurity();
         FreeStyleProject p = r.createFreeStyleProject();
         p.getPublishersList().add(new TestGroovyRecorder(
                 new SecureGroovyScript("jenkins.model.Jenkins.instance", true, null)));
+        p.getPublishersList().add(new TestGroovyRecorder(
+                new SecureGroovyScript("println(jenkins.model.Jenkins.instance.getLabels())", false, null)));
         r.assertLogNotContains("org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: "
                         + "Scripts not permitted to use staticMethod jenkins.model.Jenkins getInstance",
+                r.assertBuildStatus(Result.SUCCESS, p.scheduleBuild2(0).get()));
+        r.assertLogNotContains("org.jenkinsci.plugins.scriptsecurity.scripts.UnapprovedUsageException: script not yet approved for use",
                 r.assertBuildStatus(Result.SUCCESS, p.scheduleBuild2(0).get()));
     }
 
--- a/src/test/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest/upgradeSmokes/scriptApproval.xml
+++ b/src/test/resources/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest/upgradeSmokes/scriptApproval.xml
@@ -1,6 +1,8 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <scriptApproval plugin="script-security@1.24">
-  <approvedScriptHashes/>
+  <approvedScriptHashes>
+    <string>fccae58c5762bdd15daca97318e9d74333203106</string>
+  </approvedScriptHashes>
   <approvedSignatures>
     <string>staticMethod jenkins.model.Jenkins getInstance</string>
   </approvedSignatures>
--- a/src/test/resources/org/jenkinsci/plugins/scriptsecurity/scripts/smoke_test.yaml
+++ b/src/test/resources/org/jenkinsci/plugins/scriptsecurity/scripts/smoke_test.yaml
@@ -2,3 +2,5 @@ security:
   scriptApproval:
     approvedSignatures:
     - method java.net.URI getHost
+    approvedScriptHashes:
+    - fccae58c5762bdd15daca97318e9d74333203106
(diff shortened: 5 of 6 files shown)


yanivpaz:
When this PR will be merged into master?

MarkRx:
I don't know the maintainers would have to merge it. The last few merges were done by @rsandell so maybe he could take a look.

jglick:
> the ability to manage approved script hashes with JCasC

Why would you ever want to do this?! Whatever use case you are trying to enable, this does not sound like the right approach.

MarkRx:
We manage everything as code with JCasC across multiple Jenkins instances. There currently is no-way to whitelist scripts apart from the UI or manually modifying xml on disk. Using the UI is a not a maintainable auditable practice and bypasses SDLC verifications. Manually modifying the XML while at the same time using JCasC would result in a conflict of managing the file.

Using a hash isn't the most user friendly (especially if no comments / commit messages say what it is for) but that's the way non-sandboxed scripts are whitelisted today with this plugin. I didn't set out to rewrite it - I just need a way to whitelist some scripts.

jglick:
If this is for Pipeline, you want to use a library. If for other stuff using Groovy, your best bet is a simple plugin offering some `public` Java methods with the `@Whitelisted` annotation.

MarkRx:
It is mostly for legacy freestyle jobs though the JCasC handles both freestyle and pipelines. I do not want to whitelist the methods because they are sensitive. I intentionally want to whitelist the scripts in a controlled manner and only the scripts.

jglick:
Yeah so the plugin with `@Whitelisted` methods is the better approach in that case.

ofir-elementor:
Hi all any update when this goes live?

philsttr:
To add one more use case to what types of scripts are needed to be approved...  We use the [extended choice parameter plugin](https://plugins.jenkins.io/extended-choice-parameter/)'s groovy script option, which requires the given script to be approved.  I don't think writing a plugin is a valid option to solve this use case..

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -701,6 +702,29 @@ public synchronized void setApprovedSignatures(String[] signatures) throws IOExc
...
+        List<String> goodScriptHashes = new ArrayList<>(scriptHashes.length);
+        Pattern sha1Pattern = Pattern.compile("^[a-fA-F0-9]{40}$");
+        for (String scriptHash : scriptHashes) {
+            if (scriptHash != null && sha1Pattern.matcher(scriptHash).matches()) {
+                goodScriptHashes.add(scriptHash);
+            } else {
+                LOG.warning("Ignoring malformed script hash: " + scriptHash);
+            }
+        }
+        approvedScriptHashes.addAll(goodScriptHashes);
```
jglick, line 718:
```suggestion
        Pattern sha1Pattern = Pattern.compile("^[a-fA-F0-9]{40}$");
        for (String scriptHash : scriptHashes) {
            if (scriptHash != null && sha1Pattern.matcher(scriptHash).matches()) {
                approvedScriptHashes.add(scriptHash);
            } else {
                LOG.warning("Ignoring malformed script hash: " + scriptHash);
            }
        }
```

?

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -701,6 +702,29 @@ public synchronized void setApprovedSignatures(String[] signatures) throws IOExc
         return aclApprovedSignatures.toArray(new String[aclApprovedSignatures.size()]);
     }
 
+    @DataBoundSetter
+    public synchronized void setApprovedScriptHashes(String[] scriptHashes) throws IOException {
+        Jenkins.getInstance().checkPermission(Jenkins.RUN_SCRIPTS);
+        approvedScriptHashes.clear();
+        List<String> goodScriptHashes = new ArrayList<>(scriptHashes.length);
+        Pattern sha1Pattern = Pattern.compile("^[a-fA-F0-9]{40}$");
```
jglick, line 710:
```suggestion
        Pattern sha1Pattern = Pattern.compile("[a-fA-F0-9]{40}");
```

should suffice since you are using `matches` not `find`.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -701,6 +702,29 @@ public synchronized void setApprovedSignatures(String[] signatures) throws IOExc
...
+        for (String scriptHash : scriptHashes) {
+            if (scriptHash != null && sha1Pattern.matcher(scriptHash).matches()) {
+                goodScriptHashes.add(scriptHash);
+            } else {
+                LOG.warning("Ignoring malformed script hash: " + scriptHash);
```
jglick, line 715:
```suggestion
                LOG.warning(() -> "Ignoring malformed script hash: " + scriptHash);
```

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/JcascTest.java
```diff
@@ -26,6 +26,9 @@ public void smokeTestEntry() throws Exception {
         String[] approved = ScriptApproval.get().getApprovedSignatures();
         assertTrue(approved.length == 1);
         assertEquals(approved[0], "method java.net.URI getHost");
+        String[] approvedScriptHashes = ScriptApproval.get().getApprovedScriptHashes();
+        assertTrue(approvedScriptHashes.length == 1);
+        assertEquals(approvedScriptHashes[0], "fccae58c5762bdd15daca97318e9d74333203106");
```
jglick, line 31: More nicely done via Hamcrest, but this is following the existing example, so OK.

src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest.java
```diff
@@ -162,15 +162,17 @@ public void malformedScriptApproval() throws Exception {
...
+    @Issue({"JENKINS-57563", "JENKINS-62708"})
+    @LocalData // Just a scriptApproval.xml that whitelists 'staticMethod jenkins.model.Jenkins getInstance' and a script printing all labels
     @Test
     public void upgradeSmokes() throws Exception {
+        configureSecurity();
         FreeStyleProject p = r.createFreeStyleProject();
         p.getPublishersList().add(new TestGroovyRecorder(
                 new SecureGroovyScript("jenkins.model.Jenkins.instance", true, null)));
-        r.assertLogNotContains("org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: "
-                        + "Scripts not permitted to use staticMethod jenkins.model.Jenkins getInstance",
```
jglick, line 173: You are removing test coverage for the sandbox here.

mszabo:
Any updates on when this feature will be released?

jglick (approved):

davidmackarill:
> Adds the ability to manage approved script hashes with JCasC. With the current UX to determine the hash you can either approve it and see what was added to the scriptApproval.xml file on disk or compute the SHA-1 of the script. 

I have been able to take the hashes from the scriptApproval and it works fine however I am unable to compute my own hash of the same groovy job script, which would be ideal for when we add new scripts we can also add the hash to the `approvedScriptHashes:` list   

```
sha1sum script.groovy
c043b08a1cce0e6142bfc1269fc8e068b8570b01

```

 The approved hash in scriptsAproval.xml is different `98a015976c7392bde86f3e35d526054b684f605f` what am i missing?

jglick:
Try a `groovy:` prefix: https://github.com/jenkinsci/script-security-plugin/blob/0e7c803456e20cb241e97aaaf0ca7e4c851d9af4/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java#L381-L391 https://github.com/jenkinsci/script-security-plugin/blob/0e7c803456e20cb241e97aaaf0ca7e4c851d9af4/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/languages/GroovyLanguage.java#L40-L42

pmn3232:
I used a sample script, tried executing it. The sample.groovy generates a different hash than what I get by manually approving the same groovy script from jenkins console.

sample.groovy generated hash - 78fc02e450ce259408eafa4cc09195bd5278536a
Jenkins UI generated hash - ec1de1e53ca0110f4472848a68a906a79b650d01

`import java.security.MessageDigest;
import java.security.NoSuchAlgorithmException;

String script = '''/**
 * print top projects by workspaces' size on current node
 */
topCnt = TOP_CNT.toInteger() // how many top jobs will be printed
/**
 * file tree size in bytes
 */
treeSize = {
  it.size() + childSize(it)
}
import java.nio.file.*
/**
 * size of directory's childs
 */ 
childSize = {d ->
  if (d.directory && ! Files.isSymbolicLink(FileSystems.default.getPath(d.path))) {
    d.listFiles().inject(0) {s, f -> 
      s + treeSize(f)
    }
  } else {
    0
  }
}
/**
 * bytes to megabytes
 */
bytes2megs = {a -> 
    Math.ceil(a / (1024 * 1024)).toInteger()
}
// OS Linux
if (WORKSPACE.contains("/")) {
    slash="/"
}
// OS Windows
else {
    slash="\\"
}
//def nodeWCRoot = new File(WORKSPACE).parentFile
def nodeWCRoot = new File(WORKSPACE).parentFile.getParentFile()
println "\nWorkspace root: $nodeWCRoot:"
def freeGB = (bytes2megs(nodeWCRoot.getFreeSpace())/1024).toInteger()
println "\nFree space: $freeGB GB\n"
def lst = []
// If additional file with list of all projects exists
projectslist=new File("$WORKSPACE$slash"+".projectslist")
if (projectslist.exists() && !projectslist.isDirectory()) {
    println "Get list of projects from $projectslist\n"
    projectslist.eachLine {
        (workspace,buildinfo)=it.tokenize('|')
        file=new File(workspace)
        lst.add([bytes2megs(treeSize(file)), buildinfo, file])
    }
}
// Otherwise take everything from root of workspaces as before
else {
    println "Get list of projects from $nodeWCRoot\n"
    nodeWCRoot.eachDir{
	    lst.add([bytes2megs(treeSize(it)), "M-D-YYYY", it.name])
    }
}
def cnt = [topCnt - 1, lst.size() - 1].min()
println "Top $topCnt by workspaces' size:\n\n" + lst.sort {a, b ->
	b[0] - a[0]      
    }[0 .. cnt].collect {
	sprintf('%5d MB \t %s \t %s', *it)
    }.join('\n')''';

String language = "groovy";
System.out.println(hash(script, language));


def bytesToHex(byte[] bytes) {
    final char[] HEX_ARRAY = "0123456789abcdef".toCharArray();
        char[] hexChars = new char[bytes.length * 2];
        for (int j = 0; j < bytes.length; j++) {
            int v = bytes[j] & 0xFF;
            hexChars[j * 2] = HEX_ARRAY[v >>> 4];
            hexChars[j * 2 + 1] = HEX_ARRAY[v & 0x0F];
        }
        return new String(hexChars);
    }

    def hash(String script, String language) {
        try {
            MessageDigest digest = MessageDigest.getInstance("SHA-1");
            digest.update(language.getBytes("UTF-8"));
            digest.update((byte) ':');
            digest.update(script.getBytes("UTF-8"));
            return bytesToHex(digest.digest());
        } catch (NoSuchAlgorithmException | UnsupportedEncodingException x) {
            throw new AssertionError(x);
        }
    }
`

mcFilat:
I'm uploading approvedScriptHashes in SHA-1 format to the server using jcasc . I can literally watch in the "jcasc view page" how when the scriptler page is activated, the latter takes all the hashes and converts them to sha-512. The problem is that after that no script is approved. All scripts have green checkmarks, but inside the launch window there is a notification that the administrator must approve the script. How to get around or fix this?

jglick:
@rsandell @dwnusbaum ^^^ in reference (I guess) to 65867aa471265a16198b92fb439782ba3554da66 https://www.jenkins.io/security/advisory/2022-11-15/#SECURITY-2564

mcFilat:
@jglick 
Thank you. Now the situation is this:
1. I make sure all scriptler scripts are approved.
2. I collect their sha-512 hashes into the jcasc config (from scriptApprove.xml )
3. After setting up, I do not need to run to the scriptApproval control window, because I can run scripts. But the yellow notifications are still hanging, and prevent non-administrators from working with the scriptler.
How to make a complete update of the settings using jcasc, if the goal is to approve the scriptler scripts?
Help me please!

PR #339, opened 2021-04-29
`Base64` whitelists did not work due to use of canonical rather than binary name

jglick:
Mistake in #309/#310.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
@@ -218,6 +218,7 @@ public abstract class EnumeratingWhitelist extends Whitelist {
         }
         abstract boolean exists() throws Exception;
         final Class<?> type(String name) throws Exception {
+            // TODO this should be more strict: binary name is required for nested classes
             return ClassUtils.getClass(name);
         }
         final Class<?>[] types(String[] names) throws Exception {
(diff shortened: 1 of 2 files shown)


dwnusbaum (approved):

daniel-beck (approved):

jglick:
This plugin might be a good candidate for JEP-229:
* no end-user functionality, nobody should look at version number
* frequent little fixes & enhancements
* history of absentee maintainers (not sure about now)

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java
```diff
@@ -218,6 +218,7 @@ private static boolean is(String thisIdentifier, String identifier) {
         }
         abstract boolean exists() throws Exception;
         final Class<?> type(String name) throws Exception {
+            // TODO this should be more strict: binary name is required for nested classes
             return ClassUtils.getClass(name);
```
dwnusbaum, line 222: Maybe we could add an assertion here so that any issues with the static whitelists will cause the plugin's tests to fail?

```suggestion
            Class<?> clazz = ClassUtils.getClass(name);
            assert clazz.getName().equals(name); // The whitelists require binary names, but ClassUtils allows various syntaxes.
            return clazz;
```

jglick: Well that would basically be the same as
```suggestion
            return Class.forName(name);
```
which will not work (AFAIK) because `name` is sometimes something like `int` or `Object[]`. We want to enforce use of binary name on the actual class name portion of the string, while still processing special syntaxes that `ClassUtils` supports. I spent a few minutes thinking about it and did not come up with a simple fix so I moved on.

car-roll (approved):

PR #344, opened 2021-05-06
JENKINS-65563: Whitelist signatures to allow some dynamism

haridsv:
This change whitelists the following to address [JENKINS-65563](https://issues.jenkins.io/browse/JENKINS-65563):
- Most of `Binding` signatures (except `removeVariable` which I couldn't figure out how to, see code comment) to allow a script to dynamically update its namespace.
- `getProperty` and `invokeMethod` to allow handling properties and method calls dynamically.
- Also includes a couple of missing `HashMap` APIs.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -15,7 +15,14 @@ staticMethod groovy.json.JsonOutput toJson java.util.UUID
 new groovy.json.JsonSlurper
 method groovy.json.JsonSlurper parseText java.lang.String
 
+new groovy.lang.Binding
+method groovy.lang.Binding getVariable java.lang.String
+method groovy.lang.Binding getVariables
 method groovy.lang.Binding hasVariable java.lang.String
+# FIXME: This fails test as non-existent method. Jenkins sandbox error for accessing this method is also strange looking, it says:
+#  Scripts not permitted to use method groovy.lang.GroovyObject invokeMethod java.lang.String java.lang.Object (groovy.lang.Binding removeVariable java.lang.String).
+#method groovy.lang.Binding removeVariable java.lang.String
+method groovy.lang.Binding setVariable java.lang.String java.lang.Object
 staticField groovy.lang.Closure DELEGATE_FIRST
 staticField groovy.lang.Closure DELEGATE_ONLY
 staticField groovy.lang.Closure OWNER_FIRST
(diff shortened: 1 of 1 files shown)


dwnusbaum (changes requested):
@haridsv Thanks for the PR, but I think that many of these methods are unsafe and we cannot allow them in the sandbox.

> getProperty and invokeMethod to allow handling properties and method calls dynamically.

In general, dynamic behavior like that is almost always going to be unsafe. That kind of stuff will need to be done in a trusted shared library or in a plugin.

haridsv:
@dwnusbaum Thank you for responding! Would you be able to give more details or point to any existing references on how one would be able to make use of them to break out of sandbox? My conclusion was from the point that `getProperty` and `invokeMethod` can already be implemented (i.e., it doesn't trip any script security rules), however, it is only useful for adding new behavior and in that process loose all the existing dynamic behavior as the corresponding `super` methods are not whitelisted. The super methods are already implicitly called so what additional security risk can there be if they are allowed to be called explicitly?

Also, does your concern extend to `Binding` signatures too?

haridsv:
@dwnusbaum I understand the risk of whitelisting `getProperty` and `invokeMethod` and so dropped them from the PR. Are you OK with the signatures on `Binding`?

dwnusbaum:
> My conclusion was from the point that getProperty and invokeMethod can already be implemented (i.e., it doesn't trip any script security rules), however, it is only useful for adding new behavior and in that process loose all the existing dynamic behavior as the corresponding super methods are not whitelisted.

For the most part the sandbox only applies to actual calls or runtime behavior, it does not try to block you from defining methods that may be unsafe. Also note that [this file](https://github.com/jenkinsci/script-security-plugin/blob/e99ba9cffb0502868b05d19ef5cd205ca7e0e5bd/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/blacklist#L7-L8) contains signatures that we would never want to allow, and `GroovyObject.getProperty` and `GroovyObject.invokeMethod` are in that file. (We should add a test that automatically fails the PR if a signature is in both files to catch this kind of issue automatically.)

> Would you be able to give more details or point to any existing references on how one would be able to make use of them to break out of sandbox?

You can take a look at all of the past security advisories related to `script-security` on https://www.jenkins.io/security/advisories/ to get an idea of things that are problematic. We don't have any documentation on how to break out of the sandbox 😄 

  > I understand the risk of whitelisting getProperty and invokeMethod and so dropped them from the PR. Are you OK with the signatures on Binding?

I am not sure; I would have to give it some thought and investigation and read through Groovy's source code to see if there are any subclasses of `Binding` that might be problematic. In general my feeling is that if something is not obviously safe it is best to just not add it to the list.

Also, even if you _can_ use Groovy language features to create complicated class hierarchies and dynamic behavior in sandboxed scripts, I would encourage you to keep your sandboxed scripts as simple as possible. If you have a complex Groovy library or script that you want to use in a Pipeline, I would consider running it as a separate script using something like `sh('groovy myGroovyScript.groovy')`, in which case you don't have to worry about the sandbox at all and can test your script totally independently from Jenkins (if your script is not in a Pipeline you might not have any other options).

haridsv:
I will open separate PR for safe signatures from this PR.

PR #345, opened 2021-05-06
Update plugin parent POM; use plugin BOM

basil:
Supercedes #338, #341, and #343. Adapts to jenkinsci/bom#505.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -2,6 +2,6 @@
   <extension>
     <groupId>io.jenkins.tools.incrementals</groupId>
     <artifactId>git-changelist-maven-extension</artifactId>
-    <version>1.0-beta-7</version>
+    <version>1.2</version>
   </extension>
 </extensions>
--- a/pom.xml
+++ b/pom.xml
@@ -1,9 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>4.16</version>
+    <version>4.18</version>
     <relativePath />
   </parent>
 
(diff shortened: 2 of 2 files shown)


bitwiseman (approved):

basil:
Thanks for the quick review! I just realized I forgot to update `git-changelist-maven-extension`, so I pushed another commit that does that. Feel free to squash merge this if desired.

dwnusbaum (approved):

res0nance (approved):

jglick (approved):

basil:
> Do you know why the build of the `sample-plugin` did not catch this?

Not sure; I just upgraded and hit the same issue as you and Mark. This patch is all I can offer at the moment.

jglick:
Can reproduce via

```diff
diff --git sample-plugin/pom.xml sample-plugin/pom.xml
index 434520e..68c7698 100644
--- sample-plugin/pom.xml
+++ sample-plugin/pom.xml
@@ -95,6 +95,11 @@
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-step-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>script-security</artifactId>
+            <scope>test</scope>
+        </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-basic-steps</artifactId>
```

as

```
[WARNING] Rule 5: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for com.github.ben-manes.caffeine:caffeine:2.8.8 paths to dependency are:
+-io.jenkins.tools.bom:sample:999999-SNAPSHOT
  +-org.jenkins-ci.plugins:script-security:1.76
    +-com.github.ben-manes.caffeine:caffeine:2.8.8
and
+-io.jenkins.tools.bom:sample:999999-SNAPSHOT
  +-io.jenkins:configuration-as-code:1.51
    +-io.jenkins.plugins:caffeine-api:2.9.1-23.v51c4e2c879c8
      +-com.github.ben-manes.caffeine:caffeine:2.9.1
]
```

Maven rather arbitrarily picks up the “closest” transitive dependency, even if it is older than something “further”. Not much we can do to defend against this generally, I think; whether you will hit this regression from https://github.com/jenkinsci/bom/pull/504 will vary according to seemingly minor details of how your POM is arranged.

jglick:
Basically we need to use https://github.com/jenkinsci/plugin-pom/blob/386ed627e6fe912493a0e535ca200619ef7e99ed/pom.xml#L536-L543 to work around the fact that the way Maven picks transitive dependency versions is ridiculous.

PR #352, opened 2021-06-13
Make `SecureGroovyScriptTest` work in PCT with Lockable Resources

basil:
See jenkinsci/bom#552. While trying to add Lockable Resources to the BOM, I started tripping on this test. The test assumes there is one `._script` element in the form, but when Lockable Resources is present there is another one earlier on the page for Lockable Resources. The test finds that one instead (it's looking for the first one) and then goes off the rails. Instead I'm having the test pick the last one rather than the first one, which in this case is the one that it's looking for. This gets the test to pass both with and without Lockable Resources present.


basil:
Any chance this could be merged so that we can add Lockable Resources to BOM compatibility runs?

timja:
@bitwiseman

basil:
@car-roll Any chance this could be merged in order to improve BOM compatibility runs?

car-roll (approved):

PR #355, opened 2021-06-25
Allow `FlowInterruptedException#getResult`

basil:
Useful for implementing the following pattern in Pipeline jobs:

```groovy
import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException

currentBuild.result = 'SUCCESS'

try {
  […] // Main
} catch (FlowInterruptedException fie) {
  currentBuild.result = fie.result.toString()
  throw fie
} catch (err) {
  currentBuild.result = 'FAILURE'
  throw err
} finally {
  […] // Something that needs access to the result, like sending a Slack message
}
```

I suggest that the example on the [the CloudBees support site](https://support.cloudbees.com/hc/en-us/articles/218554077-how-to-set-current-build-result-in-pipeline) be updated as described above. This version is better because it allows aborted builds to propagate their `ABORTED` status to the notification.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,6 +73,11 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-step-api</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>io.jenkins.plugins</groupId>
       <artifactId>caffeine-api</artifactId>
     </dependency>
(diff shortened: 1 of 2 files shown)


jglick:
The whitelist addition seems harmless (other than the test failure; IIRC only the test classpath is consulted for possible matches), but what is the actual use case here? You should not be setting `currentBuild.result`. The KB article does not explain why this would ever be a good idea.

basil:
> what is the actual use case here?

The use case is sending email or Slack notifications that include the job result. With classic job types these notifications would be instances of `Notifier` or `Publisher` (with `Publisher#needsToRunAfterFinalized` overridden to return true). With Pipeline these are just regular steps that don't have access to the job result while the job is running, unless you use the template I gave above.

jglick:
That is what the `catchError` step is for, in Scripted syntax. (Declarative Pipeline offers a built-in equivalent with `post` blocks.)

basil:
Had never heard of that step before, but yeah I guess so. So I guess this PR should be closed and the CloudBees article should be updated to say something to the effect of "don't do this; use the `catchError` step instead.

PR #356, opened 2021-07-16
[JENKINS-41797] Resolve infinite loop causes by Concurrent Write to ProxyWhitelist delegates

Dohbedoh:
[JENKINS-41797](https://issues.jenkins.io/browse/JENKINS-41797)

* Created a unit test to reproduce the problem
* Fix it with `Collections.synchronizedMap(...)`

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
@@ -33,7 +33,6 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -139,12 +138,12 @@ public class ProxyWhitelist extends Whitelist {
                     this.delegates.add(delegate);
                 }
             }
-            for (ProxyWhitelist proxyWhitelist : wrappers.keySet()) {
-                proxyWhitelist.reset();
+            for (ProxyWhitelist pw : wrappers.keySet()) {
+                pw.reset();
             }
             if (this.wrappers.isEmpty()) {  // Top-level ProxyWhitelist should be the only cache
                 // Top-level ProxyWhitelist should precache the statically permitted signatures
-                ((EnumeratingWhitelist) (this.delegates.get(0))).precache();
+                ((EnumeratingWhitelist)(this.delegates.get(0))).precache();
             }
         } finally {
             writer.unlock();
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
@@ -75,16 +75,15 @@ public class ProxyWhitelist extends Whitelist {
         reset(delegates);
     }
 
-    /**
-     * This should only be called from {@link ProxyWhitelist#reset(Collection)}. So we can assume that a write 
-     * lock is held.
-     * @param wrapper the wrapper
-     */
     private void addWrapper(ProxyWhitelist wrapper) {
+        // This method should only be called from {@link ProxyWhitelist#reset(Collection)}
+        // where this thread already holds a write lock on the wrapper.
         assert wrapper.lock.writeLock().isHeldByCurrentThread();
         lock.writeLock().lock();
         try {
             wrappers.put(wrapper, null);
+            // The rest of the method only reads the current instance's fields
+            // So we downgrade this lock from write to read to reduce contention
             lock.readLock().lock();
         } finally {
             lock.writeLock().unlock();
@@ -92,10 +91,11 @@ public class ProxyWhitelist extends Whitelist {
         try {
             for (Whitelist subdelegate : delegates) {
                 if (subdelegate instanceof EnumeratingWhitelist) {
-                    ((EnumeratingWhitelist) subdelegate).clearCache();  // We only care about top-level cache
-                    continue; // this is handled specially
+                    // Discard any cache that is not the top-level cache
+                    ((EnumeratingWhitelist) subdelegate).clearCache();
+                } else {
+                    wrapper.delegates.add(subdelegate);
                 }
-                wrapper.delegates.add(subdelegate);
             }
             wrapper.methodSignatures.addAll(methodSignatures);
             wrapper.newSignatures.addAll(newSignatures);
@@ -107,33 +107,32 @@ public class ProxyWhitelist extends Whitelist {
         }
     }
 
-    /**
-     * This only exists for consistency. Access to originalDelegates should require a read lock and this 
-     * method should be called by a <b>wrapped</b> instance's {@link ProxyWhitelist#reset(Collection)}.
-     */
-    private void reset() {
+    public final void reset(Collection<? extends Whitelist> delegates) {
         lock.writeLock().lock();
+
         try {
-            reset(originalDelegates);
+            // store the incoming delegates for use during this reset
+            // and during future reset propagation
+            originalDelegates = delegates;
+            reset();
         } finally {
             lock.writeLock().unlock();
         }
     }
 
-    public final void reset(Collection<? extends Whitelist> delegates) {
-        ReentrantReadWriteLock.WriteLock writer = lock.writeLock();
-        writer.lock();
+    private void reset() {
+        lock.writeLock().lock();
 
         try {
-            originalDelegates = delegates;
-            this.delegates.clear();
+            delegates.clear();
             methodSignatures.clear();
             newSignatures.clear();
             staticMethodSignatures.clear();
             fieldSignatures.clear();
             staticFieldSignatures.clear();
 
-            this.delegates.add(new EnumeratingWhitelist() {
+            // Make the first delegate an adapter that points the fields of this proxy instance
+            delegates.add(new EnumeratingWhitelist() {
                 @Override protected List<EnumeratingWhitelist.MethodSignature> methodSignatures() {
                     return methodSignatures;
                 }
@@ -150,7 +149,7 @@ public class ProxyWhitelist extends Whitelist {
                     return staticFieldSignatures;
                 }
             });
-            for (Whitelist delegate : delegates) {
+            for (Whitelist delegate : originalDelegates) {
                 if (delegate instanceof EnumeratingWhitelist) {
                     EnumeratingWhitelist ew = (EnumeratingWhitelist) delegate;
                     methodSignatures.addAll(ew.methodSignatures());
@@ -167,15 +166,18 @@ public class ProxyWhitelist extends Whitelist {
                     this.delegates.add(delegate);
                 }
             }
-            for (ProxyWhitelist pw : wrappers.keySet()) {
-                pw.reset();
+            for (ProxyWhitelist wrapper : wrappers.keySet()) {
+                wrapper.reset();
             }
-            if (this.wrappers.isEmpty()) {  // Top-level ProxyWhitelist should be the only cache
-                // Top-level ProxyWhitelist should precache the statically permitted signatures
-                ((EnumeratingWhitelist)(this.delegates.get(0))).precache();
+            if (wrappers.isEmpty()) {
+                // The first delegate is always an adapter pointing the fields of this proxy instance
+                EnumeratingWhitelist adapter = (EnumeratingWhitelist)delegates.get(0);
+                // Top-level ProxyWhitelist should be the only cache
+                // and should precache the statically permitted signatures
+                adapter.precache();
             }
         } finally {
-            writer.unlock();
+            lock.writeLock().unlock();
         }
     }
 
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
@@ -34,6 +34,7 @@ import org.jenkinsci.plugins.scriptsecurity.sandbox.Whitelist;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


Dohbedoh:
First build definitely reproduce the problem. Now pushing the fix with `Collections.synchronizedMap(...)` to confirm that it fixes it.

Dohbedoh:
It fixes it. Now need someone to review this and confirm whether this can have non negligeable negative impact on performance.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelistTest.java
```diff
@@ -66,4 +73,31 @@ public void caching() throws Exception {
...
+     * Test concurrent modification of delegates when initializing a ProxyWhitelist. This may cause concurrent threads 
+     * to enter an infinite loop when using a {@link java.util.WeakHashMap} to hold the delegates as it is not 
+     * synchronized. This is a rare case that is also related to Garbage Collection.
+     * If contention is acceptable and/or the delegates collection behave, this test should be quite fast, just a few 
+     * seconds. A timeout is allowed to process the creation of 1000000 {@link ProxyWhitelist} using an original 
+     * delegate. If the tasks have not completed by then, we can assume that there is a problem.
+     */
+    @Issue("JENKINS-41797")
+    @Test(timeout = 30000)
+    public void testConcurrent() throws InterruptedException, IOException {
```
bitwiseman, line 86: Did this test reproduce the issue before you made the product change?

Dohbedoh: It did yes. In the first build: https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fscript-security-plugin/detail/PR-356/1/tests/.
Then the fix kicked off build 2 and everything passes

jglick:
Amending #106?

jglick:
Using `synchronizedMap` _and_ `synchronized`? Does not smell right.

jglick (commented):
Or `lock` rather.

jglick (approved):
The inconsistency is a bit weird, but seems harmless, so if the test passes now 👍

dwnusbaum (commented):
Relates to https://github.com/jenkinsci/script-security-plugin/pull/218.

Down on line 142 (can't comment on it), we iterate over `wrappers.keySet` without synchronizing on `wrapper`, which the Javadoc for `Collections.synchronizedMap` warns not to do. Do we need to add additional synchronization there?

In general I would recommend testing changes to this code on a real controller for a few days before releasing them since it's not easy to reason about the different layers of synchronization here and what the impact of any changes on real-world performance might be.

Dohbedoh:
Right. Need to synchronize the map for iteration. Will add that.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -138,12 +140,15 @@ public final void reset(Collection<? extends Whitelist> delegates) {
                     this.delegates.add(delegate);
                 }
             }
-            for (ProxyWhitelist pw : wrappers.keySet()) {
-                pw.reset();
+            Iterator<ProxyWhitelist> pwIterator = wrappers.keySet().iterator();
+            synchronized (wrappers) {
```
jglick, line 144: So this is now inside `lock.writeLock()`, and the wrappers will in turn acquire their own r/w lock and then monitor. Seems risky. I guess there should not be a deadlock so long as there are no cycles in `wrappers`, which logically there should not be…

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -138,12 +140,15 @@ public final void reset(Collection<? extends Whitelist> delegates) {
                     this.delegates.add(delegate);
                 }
             }
-            for (ProxyWhitelist pw : wrappers.keySet()) {
-                pw.reset();
+            Iterator<ProxyWhitelist> pwIterator = wrappers.keySet().iterator();
```
jglick, line 143: I think this would need to be inside the `synchronized` block.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -66,7 +68,7 @@
     private final List<EnumeratingWhitelist.FieldSignature> staticFieldSignatures = new ArrayList<EnumeratingWhitelist.FieldSignature>();
 
     /** anything wrapping us, so that we can propagate {@link #reset} calls up the chain */
-    private final Map<ProxyWhitelist,Void> wrappers = new WeakHashMap<ProxyWhitelist,Void>();
+    private final Map<ProxyWhitelist,Void> wrappers = Collections.synchronizedMap(new WeakHashMap<ProxyWhitelist,Void>());
```
jglick, line 71: Seems clearer to revert this and explicitly acquire a monitor (`synchronized (wrappers) {…}`) around every access. Or just use `lock` for this.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
Dohbedoh: `keySet()` access is synchronized and I was following the documentation:

```
It is imperative that the user manually synchronize on the returned map when traversing any of its collection views via Iterator, Spliterator or Stream:
        Map m = Collections.synchronizedMap(new HashMap());
            ...
        Set s = m.keySet();  // Needn't be in synchronized block
            ...
        synchronized (m) {  // Synchronizing on m, not s!
            Iterator i = s.iterator(); // Must be in synchronized block
            while (i.hasNext())
                foo(i.next());
        }
```

But I guess it does not hurt to add it.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
Dohbedoh: That's my understanding as well that as long as there is no cycle (and I can't find a way to produce one) it should be fine.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
Dohbedoh: I thought about this. But if using the `lock`, access to the `wrappers` would be locked for the whole reset and I thought this could have a greater impact than granular synchronization / or using thread safe collection ?

When using `lock` and externalizing [pw.wrappers.put(this, null);](https://github.com/jenkinsci/script-security-plugin/blob/script-security-1.77/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java#L123) in a private method such as:

```
    private void addWrapper(ProxyWhitelist key, Void value) {
        lock.writeLock().lock();
        try {
            this.wrappers.put(key, value);
        } finally {
            lock.writeLock().unlock();
        }
    }
```

The test passes too. And time difference is negligible, oscillates around 2s, like with the `synchronizedMap` solution. If I change the test settings either increase the number of tasks or threads in the pool, time are equivalent between both solutions.

Dohbedoh:
As recommended by @jglick, used `lock` to prevent concurrent modification.

Dohbedoh:
As recommended by @jglick, used `lock` to prevent concurrent modification.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -120,7 +120,7 @@ public final void reset(Collection<? extends Whitelist> delegates) {
                     ew.clearCache();
                 } else if (delegate instanceof ProxyWhitelist) {
                     ProxyWhitelist pw = (ProxyWhitelist) delegate;
-                    pw.wrappers.put(this, null);
+                    pw.addWrapper(this, null);
```
bitwiseman, line 123: I think 124 through 130  need to be covered by read lock  on  `delegate.lock`.  
We should take a write lock in delegate here and then downgrade to a read lock for the following lines.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -75,8 +75,51 @@ public ProxyWhitelist(Collection<? extends Whitelist> delegates) {
...
+    /**
+     * This should only be called from {@link ProxyWhitelist#reset(Collection)}. So we can assume that a write 
+     * lock is held.
+     * @param wrapper the wrapper
+     */
+    private void addWrapper(ProxyWhitelist wrapper) {
+        assert wrapper.lock.writeLock().isHeldByCurrentThread();
+        lock.writeLock().lock();
+        try {
+            wrappers.put(wrapper, null);
```
bitwiseman, line 87: This is what cause the problem - the write lock being held was on the wrapper not the delegate.  
This line needs a write lock on `this` (the delegate), the rest of the method only needs a read lock.

bitwiseman (approved):

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -75,8 +75,51 @@ public ProxyWhitelist(Collection<? extends Whitelist> delegates) {
...
     private void reset() {
-        reset(originalDelegates);
+        Collection<? extends Whitelist> delegatesCollection;
+        this.lock.readLock().lock();
+        try {
+            delegatesCollection = originalDelegates;
+        } finally {
+            lock.readLock().unlock();
+        }
+        reset(delegatesCollection);
```
dwnusbaum, line 122: `reset(Collection)` already acquires a write lock internally, so IMO it would be clearer to acquire the write lock here before calling the method rather than introducing a temporary read lock just to reassign a variable. That way you avoid consistency issues with something else possibly modifying `delegatesCollection`/`originalDelegates` (which are the same object) after the read lock here is dropped but before the write lock in `reset(Collection)` is acquired: 

```suggestion
        lock.writeLock().lock();
        try {
            reset(originalDelegates);
        } finally {
            lock.writeLock().unlock();
        }
```

bitwiseman: Ah, reentrant lock. Sure, that makes sense.

bitwiseman: Sound good but we have to check if we are still holding the write lock.
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantReadWriteLock.WriteLock.html#unlock()

```suggestion
        lock.writeLock().lock();
        try {
            reset(originalDelegates);
        } finally {
            if (lock.writeLock().isHeldByCurrentThread()) {
                lock.writeLock().unlock();
            }
        }
```

dwnusbaum: I think it's ok without that check, since `lock` increases a counter, and `unlock` decreases the counter and only actually releases the lock if the counter is 0. In this case we'll acquire the lock in `reset()`, the count will go up 1 and then down 1 in `reset(Collection)`, and then the count will go back down to 0 and be released in the finally block in `reset()`. I might be misunderstanding your concern though.

bitwiseman: Ah, I see. Agreed.

bitwiseman: It is the middle of the night for Allan, I'm going to commit Devin's suggestion to keep us moving forward.

dwnusbaum (approved):
The changes seem fine to me. I would still recommend testing this on a real instance for at least a few days before releasing it in case there are unforseen deadlocks or noticeable performance issues.

bitwiseman:
@dwnusbaum 
Allan has done a bunch of manual testing and seen no issues.

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -75,8 +75,49 @@ public ProxyWhitelist(Collection<? extends Whitelist> delegates) {
         reset(delegates);
     }
 
+    /**
+     * This should only be called from {@link ProxyWhitelist#reset(Collection)}. So we can assume that a write 
+     * lock is held.
```
jglick, line 80: maybe clearer as part of the `assert`

src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/ProxyWhitelist.java
```diff
@@ -75,8 +75,49 @@ public ProxyWhitelist(Collection<? extends Whitelist> delegates) {
...
+     * @param wrapper the wrapper
+     */
+    private void addWrapper(ProxyWhitelist wrapper) {
+        assert wrapper.lock.writeLock().isHeldByCurrentThread();
+        lock.writeLock().lock();
+        try {
+            wrappers.put(wrapper, null);
+            lock.readLock().lock();
+        } finally {
+            lock.writeLock().unlock();
```
jglick, line 90: This is now officially over my head! https://en.wikipedia.org/wiki/Software_transactional_memory where are you

bitwiseman: @jglick 

Downgrading `this.lock` from write lock to read lock to reduce contention? Do we need to comment it? 
```suggestion
            // The rest of the method only reads the current instance's fields
            // So we downgrade this.lock from write locked to read locked to reduce contention
            lock.readLock().lock();
        } finally {
            lock.writeLock().unlock();
```

Would it be better to extend @dwnusbaum 's feedback about taking write lock to this? Then we'd eagerly take write lock on each wrapper and keep it since it will be taken later when calling `reset()` on [lines 170-172](https://github.com/jenkinsci/script-security-plugin/pull/356/files#diff-b466c3e3e537c5a2de00bc55bcfa18465a442ea37ac3d92a7f2cd9ded4221f7dR170-R172).

bitwiseman:
I'm going to go ahead and merge and release this.  Feedback from testing so far has not found any issues.

jglick:
Remember to set category labels before merging.

PR #365, opened 2021-09-26
JENKINS-66735: Whitelist some common signatures and additional signatures from `CauseOfInterruption`

haridsv:
The change whitelists a few signatures listed in [JENKINS-66735](https://issues.jenkins.io/browse/JENKINS-66735)

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/jenkins-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/jenkins-whitelist
@@ -35,4 +35,4 @@ method hudson.scm.EditType getName
 method hudson.tools.ToolInstallation getHome
 method hudson.tools.ToolInstallation getName
 method jenkins.model.CauseOfInterruption getShortDescription
-method jenkins.model.CauseOfInterruption.UserInterruption getUserId
+method jenkins.model.CauseOfInterruption$UserInterruption getUserId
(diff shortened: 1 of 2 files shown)


jglick (approved):
All look harmless.

haridsv:
@jglick Thank you for the approval. I just added one more HashMap constructor, I hope that doesn't change your stance on approval.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/jenkins-whitelist
```diff
@@ -34,3 +34,5 @@ field hudson.scm.EditType EDIT
 method hudson.scm.EditType getName
 method hudson.tools.ToolInstallation getHome
 method hudson.tools.ToolInstallation getName
+method jenkins.model.CauseOfInterruption getShortDescription
+method jenkins.model.CauseOfInterruption.UserInterruption getUserId
```
jglick, line 38: Did you actually test this? I think you meant
```suggestion
method jenkins.model.CauseOfInterruption$UserInterruption getUserId
```

haridsv: Thanks a lot for catching the typo. I didn't start making the changes that need these whitelists yet, but I was expecting the test to fail for any typos so with the test passing, I didn't think that it needed to be tested. Could there be a flaw in the test?

jglick: https://github.com/jenkinsci/script-security-plugin/blob/b4f85af4477cfa0ee27a03ed0baf439788fb4c6f/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelistTest.java#L105 is supposed to be catching this. I will need to take a look.

jglick: Ah right. https://github.com/jenkinsci/script-security-plugin/blob/c3d9d689368dcb6b7ae5cfdba4c7537fc8680194/src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/EnumeratingWhitelist.java#L221

PR #387, opened 2022-02-22
Do not use Guava's `NullOutputStream`

raul-arabaolaza:
Its location has been changed in Guava 14 meaning the tests of this plugin work fine with current baseline but fail with the PCT on the upcoming lts or when baseline is updated.

I could also wait for `2.332.1` to be released and then change the baseline and update the related Guava code, but I believe is simpler to just get rid of the related Guava usage in the test given how simple it is.

*This does not affect production code* but breaks any pipeline trying to run the PCT or the tests of this plugin with newer core versions

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelectorTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/GroovyCallSiteSelectorTest.java
@@ -24,13 +24,14 @@
 
 package org.jenkinsci.plugins.scriptsecurity.sandbox.groovy;
 
-import com.google.common.io.NullOutputStream;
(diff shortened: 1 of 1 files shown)


alecharp (approved):
code is equivalent, looks good to me.

dwnusbaum (approved):

basil:
Another alternative to Guava's `com.google.common.io.NullOutputStream` is `org.apache.commons.io.output.NullOutputStream`.

jglick:
@dwnusbaum remember you need to add a category label before merging for Release Drafter, and to trigger CD if appropriate.

jglick:
Or just `ByteArrayOutputStream` discarding the result.

jglick:
> you need to add a category label before merging

If you forget, you can add the label later and manually trigger the workflow on `master`.

PR #416, opened 2022-05-07
Add script listener to track usage

meiswjn:
This PR relates to https://github.com/jenkinsci/jenkins/pull/6539 and https://issues.jenkins.io/browse/JENKINS-68404. It serves the purpose to track potentially dangerous usages of groovy scripts.
The script listener is called when a script is considered to be allowed to use. This also means that scripts running in a sandbox are not logged.


meiswjn:
@daniel-beck https://github.com/jenkinsci/jenkins/pull/6539#issuecomment-1117376441

Let me know what you think.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -467,9 +468,32 @@ public synchronized String using(@NonNull String script, @NonNull Language langu
...
+    }
+    /**
+     * Called when a script is about to be used (evaluated).
+     * @param script a possibly unapproved script
+     * @param language the language in which it is written
+     * @param run the run executing the groovy script.
+     * @return {@code script}, for convenience
+     * @throws UnapprovedUsageException in case it has not yet been approved
+     */
+    public synchronized String using(@NonNull String script, @NonNull Language language, @NonNull Run run) throws UnapprovedUsageException {
```
daniel-beck, line 482: How can `run` be `@NonNull`? Wouldn't that require all scripts to necessarily run in the context of a build?

daniel-beck: Not properly released, but https://github.com/daniel-beck/sandboxed-script-console-plugin demonstrates it's definitely doable.

https://plugins.jenkins.io/scriptler/ has a dependency on Script Security, so there's probably a sandboxed script execution outside builds there as well.

meiswjn: Fixed in e69d7cb

meiswjn: Thanks for the explanation.

I decided to make the script origin a String and not a Run.

One thing that bugs me a bit is that possible implementations must pass the source as a binding variable instead of a parameter. What do you think: Should it be a String parameter instead?

meiswjn: Decided to keep it like this as it is the best solution for now where the endpoint is only (actively) used in combination with builds. It is compatible with other approaches so I think it is fine for now.

pom.xml
```diff
@@ -16,7 +16,7 @@
   <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
-    <jenkins.version>2.332.1</jenkins.version>
+    <jenkins.version>2.355-SNAPSHOT</jenkins.version>
```
daniel-beck, line 19:
```suggestion
    <!-- TODO https://github.com/jenkinsci/jenkins/pull/6539 -->
    <jenkins.version>2.360-rc32608.d30108d0ed5a_</jenkins.version>
```

This will make the CI build work.

daniel-beck (commented):
`#configuring` might be a good alternative to `#using` because that can determine who the user actually submitting the script is. It also has `ApprovalContext` which would probably remove the need for the API change.

daniel-beck:
A basic pipeline:

```
Script: 'echo "hello world"' from origin: 'N/A' by user: 'null'
```

Once we've finalized the API here, looks like this needs a downstream change in `workflow-cps` or so.

daniel-beck:
Looks like this PR needs to be re-filed from an origin branch, or from the fork of a committer (e.g. me), to pick up the changes to the Jenkinsfile, so we have an incremental build of this.

lemeurherve:
@daniel-beck I've replayed it with the updated Jenkinsfile.

daniel-beck:
> @daniel-beck I've replayed it with the updated Jenkinsfile.

Cheater! (Thanks 😄)

meiswjn:
With https://github.com/jenkinsci/jenkins/pull/7056 being merged, I will continue on this PR soon :)

jglick:
@meiswjn are you still planning to work on this?

meiswjn:
> @meiswjn are you still planning to work on this?

Since I would love to see this feature, yes. However, there are many other more pressing things right now, but I definitely want to do this. However, if someone stumbles upon this before I find time, feel free!

PR #453, opened 2022-09-05
Adapt generic-whitelist to handle addition of String#stripIndent method in Java 15

steven-aerts:
`stripIndent` eclipses the `DefaultGroovyMethods` which is already in the generic-whitelist.
Scripts already using this method were breaking when the JVM was upstepped as it would choose standard implementation over the one provided by groovy.

Looking at git history it is not clear to me if you require a jira ticket.  If so let me know the I create one.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelistTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelistTest.java
@@ -120,8 +120,9 @@ public class StaticWhitelistTest {
                     "java.util.Iterator", "groovy.lang.Closure"),
             // Overrides CharSequence.isEmpty in Java 15+.
             new MethodSignature(String.class, "isEmpty"),
-            // Does not exist until Java 15.
+            // Do not exist until Java 15.
             new MethodSignature(CharSequence.class, "isEmpty"),
+            new MethodSignature(String.class, "stripIndent"),
             // Override the corresponding RandomGenerator methods in Java 17+.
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelistTest.java
```diff
@@ -122,6 +122,8 @@ static void sanity(URL definition) throws Exception {
             new MethodSignature(String.class, "isEmpty"),
             // Does not exist until Java 15.
             new MethodSignature(CharSequence.class, "isEmpty"),
+            // Overrides DefaultGroovyMethods.stripIndent(String) in Java 15+.
+            new MethodSignature(String.class, "stripIndent"),
```
dwnusbaum, line 126: We are using "override" here to refer specifically to Java method overriding, which is not quite what is happening in this case, so I would just note that the new method does not exist until Java 15+ like `CharSequence.isEmpty()` instead.

```suggestion
            // Do not exist until Java 15.
            new MethodSignature(CharSequence.class, "isEmpty"),
            new MethodSignature(String.class, "stripIndent"),
```

dwnusbaum (commented):
Thanks for the PR!

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/StaticWhitelistTest.java
steven-aerts: Done

dwnusbaum (approved):

PR #457, opened 2022-10-19
[JENKINS-69899] Do not visit EmptyExpression when transforming fields declared using @Field

dwnusbaum:
See [JENKINS-69899](https://issues.jenkins.io/browse/JENKINS-69899) and https://github.com/jenkinsci/groovy-sandbox/pull/94.

- [ ] ~Ensure you have provided tests - that demonstrates feature works or fixes the issue~ Tests are in groovy-sandbox

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@
     <dependency>
       <groupId>org.kohsuke</groupId>
       <artifactId>groovy-sandbox</artifactId>
-      <version>1.31-rc227.549f6a_a_e591b_</version> <!-- TODO: https://github.com/jenkinsci/groovy-sandbox/pull/94 -->
+      <version>1.31</version>
       <exclusions>
         <exclusion>
           <groupId>org.codehaus.groovy</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@
     <dependency>
       <groupId>org.kohsuke</groupId>
       <artifactId>groovy-sandbox</artifactId>
-      <version>1.30</version>
+      <version>1.31-rc227.549f6a_a_e591b_</version> <!-- TODO: https://github.com/jenkinsci/groovy-sandbox/pull/94 -->
       <exclusions>
         <exclusion>
           <groupId>org.codehaus.groovy</groupId>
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptor.java
```diff
@@ -418,7 +418,7 @@ private interface Rejector {
                 Checker.preCheckedCast(field.getType(), value, false, false, false);
                 return super.onSetAttribute(invoker, receiver, attribute, value);
             } else {
-                throw StaticWhitelist.rejectField(field);
+                rejector = () -> StaticWhitelist.rejectField(field);
```
dwnusbaum, line 421: Not related to JENKINS-69899, but fixes a minor issue in https://github.com/jenkinsci/script-security-plugin/commit/85d16bd851b355458f6396dbc4f92c3e0eebcd72. See https://github.com/jenkinsci/script-security-plugin/pull/298#discussion_r999655554 for details.

I am fixing it here in case we end up backporting the JENKINS-69899 fix, since it would be nice to backport this as well. That said, the cases it fixes were already broken before https://github.com/jenkinsci/script-security-plugin/commit/85d16bd851b355458f6396dbc4f92c3e0eebcd72, so it is not particularly important).

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -911,6 +911,8 @@ public static void assertRejected(Whitelist whitelist, String expectedSignature,
             // GroovyRuntimeException.
             if (x.getCause() instanceof RejectedAccessException) {
                 errors.checkThat(x.getMessage(), ((RejectedAccessException)x.getCause()).getSignature(), is(expectedSignature));
+            } else {
+                errors.addError(x);
```
dwnusbaum, line 915: This also fixes a minor test-only issue in https://github.com/jenkinsci/script-security-plugin/commit/85d16bd851b355458f6396dbc4f92c3e0eebcd72. We would also want this backported to avoid introducing broken tests.

pom.xml
```diff
@@ -62,7 +62,7 @@
     <dependency>
       <groupId>org.kohsuke</groupId>
       <artifactId>groovy-sandbox</artifactId>
-      <version>1.30</version>
+      <version>1.31</version>
```
jglick, line 65: BTW I think it would be a good investment of time to inline the lib into this plugin repo.

jglick (approved):

PR #459, opened 2022-10-26
Workaround for JDK-8231454 on Java 11 is no longer necessary as of 11.0.17

basil:
Applies the change from https://github.com/jenkinsci/workflow-cps-plugin/pull/600 to the copypasta in this plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
   <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
-    <jenkins.version>2.332.1</jenkins.version>
+    <jenkins.version>2.346.1</jenkins.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -16,7 +16,7 @@
   <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
-    <jenkins.version>2.332.1</jenkins.version>
+    <jenkins.version>2.346.1</jenkins.version>
```
basil, line 19: First version of core that contains https://github.com/jenkinsci/lib-version-number/pull/41, which I am using in this code.

jglick (approved):
Sure, or just unconditionally delete the workaround for a bug which IIUC does not occur on supported Java versions.

basil:
https://github.com/jenkinsci/workflow-cps-plugin/pull/600#issuecomment-1292624997

PR #493, opened 2023-04-17
Add enum.valueOf signature to the whitelist

dasanjan1296:
- [X] Please describe what you did

This will whitelist the enum.valueOf signature. Currently, I'm having to use a workaround as follows:
```
for (Chores chore : Chores.values()) {
    if (chore.name() == bom.config.chore_type) {
        chore_type = chore
    }
}
```
instead of a simple: 
```
chore_type = Chores.valueOf(bom.config.chore_type)
```

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -116,6 +116,7 @@ new java.lang.Boolean java.lang.String
 staticMethod java.lang.Boolean parseBoolean java.lang.String
 staticMethod java.lang.Boolean valueOf boolean
 staticMethod java.lang.Boolean valueOf java.lang.String
+staticMethod java.lang.Enum valueOf java.lang.Class java.lang.String
 method java.lang.CharSequence charAt int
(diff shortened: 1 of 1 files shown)


dwnusbaum (changes requested):
Someone needs to think about enum behavior in the sandbox more generally before approving this. Currently each individual enum value must be approved, see for example https://github.com/jenkinsci/script-security-plugin/blob/e8ba1732da2dd7dadce7034c5576e8f802cb230b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist#L779-L782. This PR would effectively allow access to all enum values in the JVM. Is that a security issue? Does it make it easier to exploit other issues? I am not sure.

Personally, I would start with trying to special-case this method in `SandboxInterceptor` for enum classes defined in sandboxed scripts (always safe) rather than allowing it generally.

If after considering enum behavior in general, we decide that allowing access to everything is fine, then in addition to allowing `Enum.valueOf` we should modify `SandboxInterceptor` so that direct access to enum fields is always allowed and all existing enum values in `generic-whitelist` can be removed. There are also various special cases in `groovy-sandbox` related to casting that I think should be removed at the same time if that change was made.

See also some related discussion in [JENKINS-71056](https://issues.jenkins.io/browse/JENKINS-71056).

PR #533, opened 2023-10-10
add ParserConfiguration#setParsers to blacklist

meiswjn:
Add ParserConfiguration#setParsers to blacklist. This function allows to run scripts in the agent-context without further approval by adding them as parsers in the global configuration
-> Privilege Escalation

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/blacklist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/blacklist
@@ -114,3 +114,6 @@ method org.jenkinsci.plugins.workflow.support.steps.build.RunWrapper getRawBuild
 # SECURITY-1754
 new org.kohsuke.groovy.sandbox.impl.Checker$SuperConstructorWrapper java.lang.Object[]
 new org.kohsuke.groovy.sandbox.impl.Checker$ThisConstructorWrapper java.lang.Object[]
+
+# Ability to define global parsers that include scripts without further approval
+method io.jenkins.plugins.analysis.warnings.groovy.ParserConfiguration setParsers java.util.List


meiswjn:
@uhafner I thought that this might make sense

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/blacklist
```diff
@@ -114,3 +114,6 @@ method org.jenkinsci.plugins.workflow.support.steps.build.RunWrapper getRawBuild
 # SECURITY-1754
 new org.kohsuke.groovy.sandbox.impl.Checker$SuperConstructorWrapper java.lang.Object[]
 new org.kohsuke.groovy.sandbox.impl.Checker$ThisConstructorWrapper java.lang.Object[]
+
+# Ability to define global parsers that include scripts without further approval
+method io.jenkins.plugins.analysis.warnings.groovy.ParserConfiguration setParsers java.util.List
```
jglick, line 119: https://github.com/jenkinsci/script-security-plugin/pull/533/checks?check_run_id=17555276490

jglick (commented):
Fine I guess, though oddly specific. What would have made you think this method would be safe to approve to begin with?

There are many thousands of signatures possible in the Java Platform and Jenkins core and various Jenkins plugins, many of which would be security risks. Admins really should never approve _anything_ unless they are Java developers with Jenkins expertise who have studied the method in detail and concluded that it could not be used maliciously (_and_ submitted a PR to this repo to add it to the default whitelist).

The blacklist is intended to capture a handful of signatures which are both especially dangerous, and seem like plausible candidates for approval that an admin might mindlessly approve otherwise. For example `GroovyObject.setProperty` might appear as a rejected signature if you try to use reflection-heavy idioms in a sandbox, and if you know nothing about Groovy internals this might sound unremarkable, but in fact it would open up a giant security hole to approve it.

Is this `ParserConfiguration.setParsers` something that users would commonly attempt to include in a `Jenkinsfile` or other sandboxed script?

meiswjn:
@jglick Thanks for your reply!

I think it should be part of this list because the plugin is used pretty often and using this function is documented here:
https://github.com/jenkinsci/warnings-ng-plugin/blob/master/doc/Documentation.md#creating-a-groovy-parser-programmatically

This may lead people to use this function.

PR #538, opened 2023-11-13
`convertDeprecatedApprovedClasspathEntriesThread` should be `transient`

jglick:
#468 was incorrect; we would never want to try to serialize this `Thread` in `scriptApproval.xml`! Also best to null it out when complete.


MarkEWaite (approved):
Makes sense to me.  Thanks!

PR #545, opened 2023-11-23
Fix reloading configuration from disk

Vlatombe:
This makes the `load()` method work properly when called after startup, in case of a reload configuration from disk for example.

Functionally, the `reconfigure()` method needs to be called implicitly when calling `load()`, because it updates a snapshot state that is used by running builds. If the persisted state has changed on disk then loaded, the snapshot needs to be updated accordingly.

Without this patch, the `ScriptApproval` state is correct, but anything that has changed since last time it was loaded is not reflected on running builds.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
@@ -534,12 +534,15 @@ public class ScriptApproval extends GlobalConfiguration implements RootAction {
             save();
         }
         // only call on subsequent load to avoid cycle
-        if (Jenkins.get().getInitLevel().compareTo(InitMilestone.COMPLETED) > 0) {
+        if (Jenkins.get().getInitLevel() == InitMilestone.COMPLETED) {
             try {
+                LOG.log(Level.FINE, "Reconfiguring ScriptApproval after loading configuration from disk");
                 reconfigure();
             } catch (IOException e) {
                 LOG.log(Level.WARNING, e, () -> "Failed to reconfigure ScriptApproval");
             }
+        } else {
+            LOG.log(Level.FINE, "Skipping reconfiguration of ScriptApproval during Jenkins startup sequence");
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest.java
```diff
@@ -176,6 +176,29 @@ public void upgradeSmokes() throws Exception {
...
+        ScriptApproval sa = ScriptApproval.get();
+
+        FreeStyleProject p = r.createFreeStyleProject();
+        p.getPublishersList().add(new TestGroovyRecorder(
+                new SecureGroovyScript("jenkins.model.Jenkins.instance", true, null)));
+        p.getPublishersList().add(new TestGroovyRecorder(
+                new SecureGroovyScript("println(jenkins.model.Jenkins.instance.getLabels())", false, null)));
+        r.assertLogNotContains("org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: "
+                        + "Scripts not permitted to use staticMethod jenkins.model.Jenkins getInstance",
+                r.assertBuildStatus(Result.SUCCESS, p.scheduleBuild2(0).get()));
```
jglick, line 192: Perhaps besides the point since you approved the whole script anyway, and if you had not, you would also get an error on `method jenkins.model.Jenkins getLabels`, but OK.

jglick (approved):

timja:
I suspect this has caused https://github.com/jenkinsci/acceptance-test-harness/issues/1444

(valid issues not just a test issue)

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -536,6 +533,40 @@ public ScriptApproval() {
         if (changed) {
             save();
         }
+        // only call on subsequent load to avoid cycle
+        if (Jenkins.get().getInitLevel() == InitMilestone.COMPLETED) {
+            try {
+                LOG.log(Level.FINE, "Reconfiguring ScriptApproval after loading configuration from disk");
+                reconfigure();
```
Vlatombe, line 540: Looks like this is what is causing issues with the tests, most specifically with the hot plugin load.

PR #546, opened 2023-11-27
fix: auto re-request approval once dismissed [JENKINS-63668]

meiswjn:
This PR resolves JENKINS-63668. If a job fails because a script is not approved, it automatically re-requests the approval.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
+++ b/src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
@@ -641,7 +641,10 @@ public class ScriptApproval extends GlobalConfiguration implements RootAction {
         }
         ConversionCheckResult result = checkAndConvertApprovedScript(script, language);
         if (!result.approved) {
-            // Probably need not add to pendingScripts, since generally that would have happened already in configuring.
+            // Usually. this method is called once the job configuration with the script is saved.
+            // If a script was previously pending and is now deleted, however, it would require to re-configure the job.
+            // That's why we call it again if it is unapproved in a running job.
+            this.configuring(script, language, ApprovalContext.create(), false);
             throw new UnapprovedUsageException(result.newHash);
         }
         return script;
--- a/src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest.java
+++ b/src/test/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApprovalTest.java
@@ -51,7 +51,11 @@ import java.util.logging.Level;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.nullValue;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 public class ScriptApprovalTest extends AbstractApprovalTest<ScriptApprovalTest.Script> {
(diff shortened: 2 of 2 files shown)


meiswjn:
Hey @daniel-beck,
Did you have any time to look at this PR already?

Thanks!
Jan

meiswjn:
Hey @daniel-beck, perhaps you have missed this. If you do not have time, perhaps @Wadeck has time?

Wadeck:
👋 Hello Jan, currently we (Daniel and me) are both lacking time however I pinged the maintainers of this plugin to see if they have some time for this PR :)

jglick:
Everything related to `ScriptApproval` should be considered basically deprecated (do not use this system), and adjusting the UX is certainly dangerous. If @daniel-beck or @Wadeck wants to review this and take responsibility for the behavior, go ahead (and ideally get added to the maintainer list).

rsandell:
There are still plenty of usage of `ScriptApproval` outside of pipeline and where sandboxing isn't really possible. Email-ext templates for example is the first that comes to mind.

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
```diff
@@ -594,15 +595,16 @@ boolean isEmpty() {
...
      * @param approveIfAdmin indicates whether script should be approved if current user has admin permissions
+     * @param ignoreAdmin indicates whether auto approval should be ignored, regardless of any configurations.
      * @return {@code script}, for convenience
      */
-    public synchronized String configuring(@NonNull String script, @NonNull Language language, @NonNull ApprovalContext context, boolean approveIfAdmin) {
+    public synchronized String configuring(@NonNull String script, @NonNull Language language, @NonNull ApprovalContext context, boolean approveIfAdmin, boolean ignoreAdmin) {
         final ConversionCheckResult result = checkAndConvertApprovedScript(script, language);
         if (!result.approved) {
-            if (!Jenkins.get().isUseSecurity() || 
+            if (!Jenkins.get().isUseSecurity() ||
```
rsandell, line 604: The boolean logic here has become too complex and unreadable, so it is very difficult to grok if the added logic is sound.

But why do you need the new parameter at all? From the description it seems like it should suffice to just set `approveIfAdmin` to `false`?

meiswjn: Hey @rsandell, thanks for the review!

I agree that it is too complex.
The reason for adding this can be found [here](https://github.com/jenkinsci/script-security-plugin/pull/546#issuecomment-1827751225).

Does this answer your question? If yes, I would improve the readbility of the boolean. If not, let's further discuss its need.

rsandell: Sorry I couldn't find @Wadeck 's comment so I am still a bit lost as to why, though I trust that there is a legitimate reeason for it I just would like to understand that reason :sweat_smile: 

If you could clean up the logic to make it more readable that would be hugely helpful :) Maybe it is just enough to reformat the code?

rsandell:
Just as an FYI I remember explicitly doing this in email-ext at least for when a template is picked up from the workspace; it will call configuring followed by using every time, which might be annoying for some admins :sweat_smile:

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
meiswjn: No worries!
https://github.com/jenkinsci/script-security-plugin/pull/300#issuecomment-685495550
There you go :) Last sentence in that comment.

What ignoreAdmin does can be seen [here](https://github.com/jenkinsci/script-security-plugin/commit/3a8f7ee16ae579933ee1f6cfaf2ca356fa4875ff#diff-5ed519b7b22dea269dd4d1feeb9ba76ced885593609c49fc25e9eb429e68702fR659). When the using() function is called from a job, it will not cause scripts to be approved automatically. If the admin did set up the script, it would have already been approved on saving the configuration.

It is supposed to prevent the scenario, where an admin starts a job, not knowing that it will approve a potentially dangerous script in it.

I am reformatting it in a sec

meiswjn:
> Just as an FYI I remember explicitly doing this in email-ext at least for when a template is picked up from the workspace; it will call configuring followed by using every time, which might be annoying for some admins 😅

What would be the outcome of this scenario? I think once the script is approved, the admin does not have to do anything unless approval is revoked :)

src/main/java/org/jenkinsci/plugins/scriptsecurity/scripts/ScriptApproval.java
meiswjn: >  though I trust that there is a legitimate reeason for it I just would like to understand that reason 😅

Yes, please! I am new to this repository's source code. There is a chance that I am completely wrong with that new boolean and since this is security relevant, I totally support that you question it.

meiswjn: I tried to make it much more readable, even though not as "fancy" as before. Does this help? @rsandell

jglick:
> outside of pipeline and where sandboxing isn't really possible. Email-ext templates for example

Getting off topic, but FWIW many cases would be better handled by replacing that plugin with e.g.

```groovy
mail subject: 'Failure', …, body: """
Build broke! $BUILD_URL
"""
```

meiswjn:
Hello @rsandell,
Given that this PR is open for several months now, would you be able to provide a timeline or delegate the review?

Thanks! I appreciate your work!

rsandell:
I am very sorry for the delay, our team is quite swamped at the moment and this keeps falling between the cracks. So thanks for the reminder!
I've set a reminder so that me or someone in my team doesn't forget about it again. So when there is a calm hour we'll take a look again.

meiswjn:
> I am very sorry for the delay, our team is quite swamped at the moment and this keeps falling between the cracks. So thanks for the reminder! I've set a reminder so that me or someone in my team doesn't forget about it again. So when there is a calm hour we'll take a look again.

Thanks, @rsandell! Would you have an ETA for me?

meiswjn:
In the current state, this PR is not compatible with some of the recent implemented changes. Moving it to draft until I find some time.

PR #549, opened 2023-12-19
Overhauled `ProxyWhitelist`

jglick:
Fixes https://github.com/jenkinsci/acceptance-test-harness/issues/1444. (I guess; I do not care to run ATH locally, but anyway the problenatic code is completely rewritten so it ought to be possible to revert https://github.com/jenkinsci/acceptance-test-harness/pull/1445.) Motivated by https://github.com/jenkinsci/script-security-plugin/pull/545#discussion_r1430436534: there was some code that attempted to work around Guice cycles, but I guess it behaved nondeterministically depending on extension load order. Simplified the whole thing, which required rewriting `ProxyWhitelist` to be much simpler and not try to cache anything, and rewriting `EnumeratingWhitelist` to use a simpler cache system (replacing #180 and various other accreted patches).

PR #555, opened 2024-02-01
Avoid race condition in `ApprovedWhitelist` reconfiguration

jglick:
https://github.com/jenkinsci/script-security-plugin/pull/549#discussion_r1439742106 apparently can cause a functional test flake in CloudBees CI where a signature is approved and then two builds requiring that signature are started nearly simultaneously; sometimes one of them misses the update. Was trying to avoid a lock here but I do not see how.


dwnusbaum (approved):

jglick:
(Waiting for an incremental build so I can test in PCT, but the Windows build machine seems extremely slow today.)

PR #565, opened 2024-04-18
whitelist replaceAll(pattern, String)

ysmaoui:
adding whitelisting for String.replaceAll(pattern, String) 

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
+++ b/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
@@ -796,7 +796,7 @@ method java.util.regex.MatchResult group int
 method java.util.regex.MatchResult group java.lang.String
 method java.util.regex.MatchResult groupCount
 method java.util.regex.MatchResult hasMatch
-method java.util.regex.MatchResult namedGroups 
+method java.util.regex.MatchResult namedGroups
 method java.util.regex.MatchResult start
 method java.util.regex.MatchResult start int
 method java.util.regex.MatchResult start java.lang.String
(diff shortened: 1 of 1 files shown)


jglick:
Note that these are not currently usable in Pipeline: https://github.com/jenkinsci/workflow-cps-plugin/blob/7141c520ff447a70876d806df9a6c159d12525c9/dgm-builder/src/main/resources/com/cloudbees/groovy/cps/tool/translatable.txt#L183-L186

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1373,8 +1373,10 @@ staticMethod org.codehaus.groovy.runtime.StringGroovyMethods readLines java.lang
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods readLines java.lang.String
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods replaceAll java.lang.CharSequence java.lang.CharSequence groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods replaceAll java.lang.CharSequence java.util.regex.Pattern groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.StringGroovyMethods replaceAll java.lang.CharSequence java.util.regex.Pattern java.lang.CharSequence
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods replaceAll java.lang.String java.lang.String groovy.lang.Closure
 staticMethod org.codehaus.groovy.runtime.StringGroovyMethods replaceAll java.lang.String java.util.regex.Pattern groovy.lang.Closure
+staticMethod org.codehaus.groovy.runtime.StringGroovyMethods replaceAll java.lang.String java.util.regex.Pattern java.lang.String
```
jglick, line 1379: This is deprecated: https://github.com/apache/groovy/blob/GROOVY_2_4_21/src/main/org/codehaus/groovy/runtime/StringGroovyMethods.java#L2613-L2620

Since `String` implements `CharSequence`, the sandbox should not be calling this.

I would recommend adding unit tests demonstrating that an entry in this whitelist is actually required.

I suspect `StaticWhitelistTest` is failing to mechanically detect useless entries.

dwnusbaum:
> Note that these are not currently usable in Pipeline: https://github.com/jenkinsci/workflow-cps-plugin/blob/7141c520ff447a70876d806df9a6c159d12525c9/dgm-builder/src/main/resources/com/cloudbees/groovy/cps/tool/translatable.txt#L183-L186

I don't think they need to be translated because they don't have a `Closure` (or `Comparator`, etc.) parameter.

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
jglick: :shrug: Either way, should be harmless to allow it.

jglick (approved):

dwnusbaum (approved):

PR #923, opened 2026-07-14
Cleaner `hashClasspathEntry` impl

jglick:
Amends 65867aa471265a16198b92fb439782ba3554da66. Supersedes #922. Rather than ignore the SB warning, it seemed clearer to write the drain of the `DigestInputStream` this way.


gbhat618 (approved):
nice 🎉

PR #928, opened 2026-09-17
Whitelist NullObject methods that got rejected because of SECURITY-3931 fix

nevingeorgesunny:
## Summary
- In https://github.com/jenkinsci/script-security-plugin/commit/f5437ae62888b44e760cd7c4ca551e1948d49163 made `SandboxInterceptor` route operations on a `null` receiver through the sandbox instead of bypassing it. Because `generic-whitelist` contained no `NullObject` entries at all, script idioms that had always worked started failing with `RejectedAccessException`.

This PR adds the `NullObject` entries needed to restore that behaviour, without weakening the null-receiver interception the advisory introduced.

### Entries added
- method org.codehaus.groovy.runtime.NullObject plus java.lang.String
- method org.codehaus.groovy.runtime.NullObject asBoolean
- method org.codehaus.groovy.runtime.NullObject asType java.lang.Class
- method org.codehaus.groovy.runtime.NullObject is java.lang.Object
- method org.codehaus.groovy.runtime.NullObject iterator
- method org.codehaus.groovy.runtime.NullObject plus java.lang.Object

## Test plan
- [x] `nullReceiverMethodsPermittedByGenericWhitelist` test passed .


src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1295,6 +1295,7 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimil
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimilarCollection java.util.Collection int
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimilarMap java.util.Map
 staticMethod org.codehaus.groovy.runtime.InvokerHelper asIterator java.lang.Object
+method org.codehaus.groovy.runtime.NullObject plus java.lang.String
```
gbhat618, line 1303: could there be similar issue with the other way round ?
```
def bar = null;
echo "foo" + bar;
```

nevingeorgesunny: let me do a quick check

nevingeorgesunny: I tried a job with 
```
try {
  def bar = null;
  echo "foo" + bar;
} catch (e) {
  echo "PLUS_EXCEPTION: ${e.class.name}: ${e.message}"
}
```

I got 

```
[Pipeline] echo
foonull
```

gbhat618 (approved):

nevingeorgesunny:
I also tested a few other null-receiver cases alongside this fix, to check whether they're affected:
```
try {
  def foo = null
  def x = foo[0]
  echo "GETAT_RESULT: ${x}"
} catch (e) {
  echo "GETAT_EXCEPTION: ${e.class.name}: ${e.message}"
}

try {
  def foo = null
  foo[0] = 1
  echo "PUTAT_RESULT: ok"
} catch (e) {
  echo "PUTAT_EXCEPTION: ${e.class.name}: ${e.message}"
}

try {
  def foo = null
  foo.trim()
  echo "TRIM_RESULT: ok"
} catch (e) {
  echo "TRIM_EXCEPTION: ${e.class.name}: ${e.message}"
}

try {
  def foo = null
  def v = foo.@bar
  echo "FIELD_RESULT: ${v}"
} catch (e) {
  echo "FIELD_EXCEPTION: ${e.class.name}: ${e.message}"
}
```
With this fix applied, I got:
```
GETAT_EXCEPTION: java.lang.NullPointerException: Cannot invoke method getAt() on null object
PUTAT_EXCEPTION: java.lang.NullPointerException: Cannot invoke method putAt() on null object
TRIM_EXCEPTION: java.lang.NullPointerException: Cannot invoke method trim() on null object
FIELD_EXCEPTION: <a MissingProperty/MissingFieldException, depending on script-security version>: no such field
```

But these are not regression as I was getting these same error before [1422.v06869826dd9b_](https://github.com/jenkinsci/script-security-plugin/releases/tag/1422.v06869826dd9b_) version also

apuig:
> I also tested a few other null-receiver cases 

What about 

```groovy
try {
    def foo = null
    foo += 1
} catch (e) {
    echo "REJECTED: ${e.message}"
}
// REJECTED: Scripts not permitted to use method org.codehaus.groovy.runtime.NullObject plus java.lang.Object

def foo = null
foo += "found ${1+1} items"
echo "ALLOWED: ${foo}"
// ALLOWED: nullfound 2 items
```
`NullObject` declares two `plus` overloads, so suggesting to also consider a whitelisting `method org.codehaus.groovy.runtime.NullObject plus java.lang.Object` to avoid a counter/accumulator pattern like  (`counts[key] += 1`, `total += someBigDecimal`) to still break

nevingeorgesunny:
> What about
 
yes thank you , good catch , i got 
 
 ```
[Pipeline] echo
REJECTED: Scripts not permitted to use method org.codehaus.groovy.runtime.NullObject plus java.lang.Object
[Pipeline] echo
ALLOWED: nullfound 2 items
```

for 

```
try {
    def foo = null
    foo += 1
} catch (e) {
    echo "REJECTED: ${e.message}"
}
// REJECTED: Scripts not permitted to use method org.codehaus.groovy.runtime.NullObject plus java.lang.Object
try {
    def foo = null
    foo += "found ${1+1} items"
    echo "ALLOWED: ${foo}"
// ALLOWED: nullfound 2 items
} catch (e) {
    echo "REJECTED: ${e.message}"
}
```

I will push a fix now

**Comment by jglick** [comment]
Some tests would be appreciated. We usually do not bother with test coverage for simple whitelist additions, but this case is more subtle than most.

**Comment by jglick** [comment]
Something in https://github.com/jenkinsci/script-security-plugin/blob/06869826dd9b3cc608d6bae6ca8db72241c45154/src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java#L430-L458 I guess, using

```java
assertEvaluate(new GenericWhitelist(), …
```

jglick:
> In a recent change

Please be specific. f5437ae62888b44e760cd7c4ca551e1948d49163 right?

nevingeorgesunny:
There's also a problem with `NullObject.plus(Object)`, `NullObject.asBoolean()`, `NullObject.asType(Class)`, `NullObject.is(Object)` and `NullObject.iterator()` — same root cause, they all resolve to real methods on NullObject and now hit the whitelist check.

Testing this locally and adding tests, will commit soon and update the PR title.

nevingeorgesunny:
> Please be specific. https://github.com/jenkinsci/script-security-plugin/commit/f5437ae62888b44e760cd7c4ca551e1948d49163 right?

done updaed the PR description 

> Some tests would be appreciated

done added tests in the latest commits

> NullObject declares two plus overloads, so suggesting to also consider a whitelisting method

done added the whitelisting in the latest commits , and also added others

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1295,6 +1295,14 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimil
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimilarCollection java.util.Collection int
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimilarMap java.util.Map
 staticMethod org.codehaus.groovy.runtime.InvokerHelper asIterator java.lang.Object
+method org.codehaus.groovy.runtime.NullObject asBoolean
+method org.codehaus.groovy.runtime.NullObject asType java.lang.Class
+staticMethod org.codehaus.groovy.runtime.NullObject getNullObject
```
jglick: Why is this here? I do not suppose any real user script was calling this?

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1295,6 +1295,14 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimil
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimilarCollection java.util.Collection int
 staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimilarMap java.util.Map
 staticMethod org.codehaus.groovy.runtime.InvokerHelper asIterator java.lang.Object
+method org.codehaus.groovy.runtime.NullObject asBoolean
```
jglick, line 1298: `return false` OK

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1295,6 +1295,14 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimil
...
+method org.codehaus.groovy.runtime.NullObject asType java.lang.Class
```
jglick, line 1299: `return null` OK

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1295,6 +1295,14 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimil
...
+staticMethod org.codehaus.groovy.runtime.NullObject getNullObject
+method org.codehaus.groovy.runtime.NullObject is java.lang.Object
```
jglick, line 1300: `other == null` OK

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1295,6 +1295,14 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimil
...
+method org.codehaus.groovy.runtime.NullObject iterator
```
jglick, line 1301: `Collections.EMPTY_LIST.iterator()` OK

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1295,6 +1295,14 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimil
...
+method org.codehaus.groovy.runtime.NullObject plus java.lang.Object
```
jglick, line 1302: calls `String.valueOf` (OK) then throws `NullPointerException` (OK)

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1295,6 +1295,14 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimil
...
+method org.codehaus.groovy.runtime.NullObject plus java.lang.String
```
jglick, line 1303: weird but basically just calls `toString()` on two objects, OK

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
```diff
@@ -1295,6 +1295,14 @@ staticMethod org.codehaus.groovy.runtime.DefaultGroovyMethodsSupport createSimil
...
+method org.codehaus.groovy.runtime.NullObject with groovy.lang.Closure
```
jglick: `DefaultGroovyMethods.with( null, closure )` this is more complicated to analyze, pull it out please, nobody asked for this anyway

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -457,6 +457,40 @@ private Unsafe() {}
...
+    @Test public void nullReceiverMethodsPermittedByGenericWhitelist() throws Exception {
+        assertEvaluate(new GenericWhitelist(), "nullbar", "def foo = null; foo += 'bar'; foo");
+        assertEvaluate(new GenericWhitelist(), "nullfound 3 items",
+                "def n = 3; def summary = null; summary += \"found ${n} items\"; summary");
+        assertEvaluate(new GenericWhitelist(), false, "def foo = null; foo.asBoolean()");
+        assertEvaluate(new GenericWhitelist(), null, "def foo = null; foo.asType(String)");
+        assertEvaluate(new GenericWhitelist(), true, "def foo = null; foo.is(null)");
+        assertEvaluate(new GenericWhitelist(), false, "def foo = null; foo.is('x')");
+        assertEvaluate(new GenericWhitelist(), false, "def foo = null; foo.iterator().hasNext()");
+        assertEvaluate(new GenericWhitelist(), "null", "def foo = null; foo.getNullObject().toString()");
```
jglick: Huh? This is a `static` method.
```suggestion
```

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
```diff
@@ -457,6 +457,40 @@ private Unsafe() {}
...
+        assertEvaluate(new GenericWhitelist(), 42, "def foo = null; foo.with { 42 }");
```
jglick: better omit please
```suggestion
```

jglick (changes requested):
Better drop `getNullObject` and `with`, to keep this focused on the clearly safe methods that scripts were likely to actually be calling.

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
nevingeorgesunny: done in https://github.com/jenkinsci/script-security-plugin/pull/928/changes/2a08701bb5ae41ede40301511ced3053aad1023e

src/test/java/org/jenkinsci/plugins/scriptsecurity/sandbox/groovy/SandboxInterceptorTest.java
nevingeorgesunny: done in https://github.com/jenkinsci/script-security-plugin/pull/928/changes/2a08701bb5ae41ede40301511ced3053aad1023e

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
nevingeorgesunny: removed in https://github.com/jenkinsci/script-security-plugin/pull/928/changes/2a08701bb5ae41ede40301511ced3053aad1023e

src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist
nevingeorgesunny: removed in https://github.com/jenkinsci/script-security-plugin/pull/928/changes/2a08701bb5ae41ede40301511ced3053aad1023e

nevingeorgesunny:
> Better drop getNullObject and with, to keep this focused on the clearly safe methods that scripts were likely to actually be calling.

dropped in https://github.com/jenkinsci/script-security-plugin/pull/928/changes/2a08701bb5ae41ede40301511ced3053aad1023e

jglick (approved):

